PR #1, opened 2019-03-19
Update setup_env.sh

dvnagorny:
Remove unnecessary settings


SDxKeeper:
Yes. These environment variables should be set by media driver install script.

PR #3, opened 2019-04-02
Update CMakeLists to build with latest DLDT

fenghaitao:
With those changes, I could build gst-video-analytics for AARCH64 with the dldt (https://github.com/fenghaitao/dldt/tree/arm) forked from https://github.com/opencv/dldt.

### DIFF
--- a/gst/elements/CMakeLists.txt
+++ b/gst/elements/CMakeLists.txt
@@ -59,7 +59,6 @@ PRIVATE
 
 target_link_libraries(${TARGET_NAME}
 PRIVATE
-        ie_cpu_extension
         ${OpenCV_LIBS}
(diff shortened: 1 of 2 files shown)


SDxKeeper:
Hi Haitao,

Thank you for your interest in gst-video-analytics product. I've reproduced your issue and checked fix you propose.

Your change to inference_backend/image_inference/openvino/CMakeLists.txt is correct

While change to the gst/elements/CMakeLists.txt can be simplified - 
`ie_cpu_extension` is not needed for this element at all, so this line can be removed.

I'll merge your request once you fix comments above.

fenghaitao:
Update the patch to address your comment.

PR #48, opened 2020-02-25
Fixed docker build script

kumarpus:
Changes the build_docker_image.sh to make it more generalize. Usually `http_proxy` used instead of `HTTP_PROXY`, I had changed the script to accept both. 
Had build issue with `--build-arg dldt=${dldt}`, after removing works fine. These changes are verified and tested, please consider them. Thanks


umed:
thank you, @kumarpus :+1:

kumarpus:
welcome :-)

PR #401, opened 2024-02-03
fix: perform postprocessing outside critical section

eslam-ahmed-khair:
In cases where the inference engine is shared among multiple streams, having a batch size > 1 and the postprocessing for the batch is time-consuming, this results in reduced performance as the time needed to form a complete batch is increased due to blocking the streamer threads trying to push more frames


tjanczak:
included as part of 2024.0 release

eslam-ahmed-khair:
> included as part of 2024.0 release

Closing this pull request as a result.

PR #418, opened 2024-05-01
fix: avoid-driver-context-switch

aminatef:
- In scenarios where multiple sources share the same inference engine, each source possessing its own VADisplay (driver context), and where the inference engine uses a different VADisplay (driver context),a conversion of VASurfaces from the source driver context to the inference engine driver context is initiated.

- This conversion involves the creation of an additional VASurface within the inference engine's driver context, followed by the transfer of the surface to it ,all on frame basis, this conversion process introduces a performance bottleneck resulting in GPU under utilization and a decrease in the number of frames submitted for inference.

- The implemented solution ensures that all inference elements  use the same VADisplay as the inference engine to avoid the performance bottleneck introduced by the conversion.

- Benchmark results showed up to 40% (from 723 FPS to 977 FPS in Yolov8s) and 100% (from 951 FPS to 1800 FPS in ResNet10) performance improvement post-fix enhancing GPU utilization and frame inference submission rate

- Benchmark results were taken at batch size equals 64 and 30 sources

### DIFF
--- a/src/monolithic/gst/inference_elements/base/inference_singleton.cpp
+++ b/src/monolithic/gst/inference_elements/base/inference_singleton.cpp
@@ -5,6 +5,8 @@
  ******************************************************************************/
 
 #include "inference_singleton.h"
+#include "gva_base_inference_priv.hpp"
+#include "inference_backend/buffer_mapper.h"
 
 #include "gva_base_inference.h"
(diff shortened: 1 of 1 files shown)


dmichalo:
Closing - outdated PR.

PR #438, opened 2024-09-12
Fix Type Mismatch in decode_inference.cpp

deshpanda:
This pull request addresses a type mismatch issue encountered while building DLStreamer from source on a host system following the [Install guide](https://dlstreamer.github.io/get_started/install/install_guide_ubuntu.html#option-3-compile-intel-dl-streamer-pipeline-framework-from-sources-on-host-system) During the `make -j` process, the following error was observed

```
/home/samyak/intel/dlstreamer_gst/samples/ffmpeg_openvino/cpp/decode_inference/decode_inference.cpp:81:87: error: invalid conversion from ‘const AVCodec**’ to ‘AVCodec**’ [-fpermissive]
   81 |         int video_stream = av_find_best_stream(input_ctx, AVMEDIA_TYPE_VIDEO, -1, -1, &codec, 0);
      |                                                                                       ^~~~~~
      |                                                                                       |
      |                                                                                       const AVCodec**
```

Upon reviewing `decode_inference.cpp`, it was identified that `AVCodec *codec` was incorrectly marked as `const`, whereas the `av_find_best_stream` function expects a non-const pointer:

```
int av_find_best_stream(AVFormatContext *ic, AVMediaType type, int wanted_stream_nb,
                        int related_stream, AVCodec **decoder_ret, int flags);
```

This update removes the `const` qualifier from `codec` to match the expected type in the function prototype, thus resolving the type mismatch error and enabling a successful build with `make -j`.

### DIFF
--- a/samples/ffmpeg_openvino/cpp/decode_inference/decode_inference.cpp
+++ b/samples/ffmpeg_openvino/cpp/decode_inference/decode_inference.cpp
@@ -77,7 +77,7 @@ int main(int argc, char *argv[]) {
         AVInputFormat *input_format = NULL;
         AVFormatContext *input_ctx = NULL;
         DLS_CHECK_GE0(avformat_open_input(&input_ctx, FLAGS_i.data(), input_format, NULL));
-        const AVCodec *codec = nullptr;
+        AVCodec *codec = nullptr;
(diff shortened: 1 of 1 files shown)


deshpanda:
Hi @nszczygl9, could you please review this pull request when you have a chance? Let me know if anything needs clarification. Thanks!

tbujewsk:
Thank you for information. This is/was a known issue already solved with the latest instruction installation.

PR #498, opened 2026-01-04
Fix libva path setup in setup_dls_env.ps1

yunowo:
There is no exact match `VideoAccelerationCompatibilityPack` in `$pathEntries`, so a new entry is added to PATH when ever the script is excuted.

### DIFF
--- a/scripts/setup_dls_env.ps1
+++ b/scripts/setup_dls_env.ps1
@@ -1,6 +1,6 @@
 #Requires -RunAsAdministrator
 # ==============================================================================
-# Copyright (C) 2025 Intel Corporation
+# Copyright (C) 2026 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
 # ==============================================================================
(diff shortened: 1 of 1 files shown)


scripts/setup_dls_env.ps1
```diff
@@ -173,7 +173,7 @@ Write-Host 'Setting variables: LIBVA_DRIVER_NAME, LIBVA_DRIVERS_PATH, Path (for
 [Environment]::SetEnvironmentVariable('LIBVA_DRIVERS_PATH', (Get-Item .).FullName + '\Microsoft.Direct3D.VideoAccelerationCompatibilityPack.1.0.2\build\native\x64\bin\', [System.EnvironmentVariableTarget]::User)
```
qianlongding: we should change line 173, usually the current directory is **C:\dlstreamer_dlls**, but this directory does not contain **Microsoft.Direct3D.VideoAccelerationCompatibilityPack.1.0.2**. The LIBVA_DRIVERS_PATH should be set to the correct path, and only then should it be added to the Path variable.

tbujewsk (approved):

PR #504, opened 2026-01-06
Remove duplicate gvametapublish element register

yunowo:
On Windows, both `gstvideoanalytics.dll` and `gvametapublish.dll` registers `gvametapublish` and `gvametapublishfile` elements and creates conflict when loading dll. Remove them from `gstvideoanalytics.dll`.
Fixes ITEP-82821.


BaoHuiling (approved):

qianlongding:
After this change, both gstvideoanalytics.dll and gstvideoanalytics.so will not include gvametapublish and gvametapublishfile

PR #583, opened 2026-02-02
[Windows CI] Add download models on Windows

yangjianfeng1208:
### Description

1. Install the opensource tool rclone on windows runner
2. Using rclone to copy models from one of the linux runners

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/.github/workflows/dls-download-models.yaml
+++ b/.github/workflows/dls-download-models.yaml
@@ -70,183 +70,182 @@ jobs:
           WIN_ARRAY=$(echo "$WIN_LABELS" | jq -Rc 'split(",")')
           echo "windows_runners=$WIN_ARRAY" >> $GITHUB_OUTPUT
 
-  # update_linux_hosts:
-  #   name: Update on Linux runners
-  #   needs: setup-runners
-  #   permissions:
-  #     contents: read
-  #   strategy:
-  #     fail-fast: false
-  #     matrix:
-  #       runner: ${{ fromJson(needs.setup-runners.outputs.linux_runners) }}
-  #   runs-on:
-  #     - ${{ matrix.runner }}
-  #   steps:
-  #     - name: Get script
-  #       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # 5.0.0
-  #       with:
-  #         persist-credentials: false
-  #         path: dlstreamer-repo
+  update_linux_hosts:
+    name: Update on Linux runners
+    needs: setup-runners
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: ${{ fromJson(needs.setup-runners.outputs.linux_runners) }}
+    runs-on:
+      - ${{ matrix.runner }}
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/dls-download-models.yaml
```diff
@@ -225,3 +239,250 @@ jobs:
         run: |
           rm -rf dlstreamer-repo .virtualenvs
           rm -rf $HOME/.virtualenvs
+
+  update_windows_hosts:
+    if: ${{ !inputs.specific_os || inputs.specific_os == 'all' || inputs.specific_os == 'windows' }}
+    name: Update on Windows runners
+    permissions:
+      contents: read
+    runs-on:
```
dmichalo, line 255: I would suggest to use the same approach as for Linux - using runner-specific labels (the same as runners names). In case of Windows, it is DLS-ARL-10 at the moment.

.github/workflows/dls-download-models.yaml
```diff
@@ -225,3 +239,250 @@ jobs:
...
+      - self-hosted
+      - dlstreamer
+      - ARL
+      - windows
+    steps:
+      - name: Get script
```
dmichalo: It fetches the entire repo in fact - maybe adjust the step's name.

.github/workflows/dls-download-models.yaml
```diff
@@ -225,3 +239,250 @@ jobs:
...
+          }
+          if (Test-Path $publicOld) {
+              Rename-Item -Path $publicOld -NewName "public"
+          }
+          if (Test-Path $publicOldOld) {
+              Rename-Item -Path $publicOldOld -NewName "public_old"
+          }
+          "Old Windows public models restored" | Out-File -FilePath $env:GITHUB_STEP_SUMMARY -Encoding utf8 -Append
+
+      - name: Download OMZ models on Windows (setup venv, dependencies and download)
```
dmichalo: No need for 'Prepare directories' for OMZ?

samples/download_public_models.sh
```diff
@@ -11,7 +11,11 @@ QUANTIZE=${2:-""} # Supported values listed in SUPPORTED_QUANTIZATION_DATASETS b
 # Save the directory where the script was launched from
 LAUNCH_DIR="$PWD"
 
-. /etc/os-release
+if [ -f /etc/os-release ]; then
```
dmichalo: Changes in this script will require some discussion probably, while it is a subject of changes right now and it will be simplified. Nevertheless, we have to test it.

yangjianfeng1208: Yes, I agreed with you, I just revert this PR to draft status.
Perhaps we can proceed this after you guys have reached some conclusions.

yunowo: Maybe refactor this script into a python file that handles model conversion (platform independent), a bash script & a powershell script for venv and pip install.

.github/workflows/dls-download-models.yaml
yangjianfeng1208: Oh, I missed this

PR #620, opened 2026-02-17
[Samples]: Add VLM alerts sample in python 1/2

oonyshch:
### Description

This PR is intended to:
- Add vlm_alerts.py sample application
- Download video if URL provided
- Export HuggingFace VLM to OpenVINO IR using optimum-cli
- Build and run GStreamer pipeline with gvagenai
- Write JSONL metadata output
- Add requirements.txt for exporter stack

Tested by running locally for all mentioned components on ARL-H.

### Checklist:

### Next steps:

- Modify the genai.cpp file to support displaying VLM inference result on the obtained video.
- Add support for multiple widths and heights of the processed video.
- Add list of suggested VLMs.


samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+        return local_path.resolve()
+
+    print(f"[video] downloading {path_or_url}")
+    request = urllib.request.Request(
+        path_or_url,
+        headers={"User-Agent": "Mozilla/5.0"},
+    )
+
+    with urllib.request.urlopen(request) as response, open(local_path, "wb") as file:
+        file.write(response.read())
```
walidbarakat: shouldn't we check if the response is ok i.e not timed out and the video is downloaded in a good format?

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(
+        description="DLStreamer VLM Alerts sample"
+    )
+    parser.add_argument("video")
+    parser.add_argument("model")
+    parser.add_argument("question")
+    parser.add_argument("--device", default="GPU")
+    parser.add_argument("--max-tokens", type=int, default=20)
+    parser.add_argument("--frame-rate", type=float, default=1.0)
```
walidbarakat: let's be consistent about the way the user pass the arguments. and make all the parameters marked with `--<arg-name>` to be easy parsed. and i'd recommend adding a help message in case the user passed the wrong set of parameters i.e one of the required parameters is missing.

and better to be more explicit with the arguments name and have something like `--video-path` and `--video-url` so it will be more conscious for the user and for the script to be clear about separation of responsibilities between functions and the same for the model

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+from typing import Tuple
+
+import gi
+gi.require_version("Gst", "1.0")
+from gi.repository import Gst, GLib # pylint: disable=no-name-in-module, wrong-import-position
+
+BASE_DIR = Path(__file__).resolve().parent
+VIDEOS_DIR = BASE_DIR / "videos"
+MODELS_DIR = BASE_DIR / "models"
+RESULTS_DIR = BASE_DIR / "results"
```
walidbarakat: good, but it would be good as well to let the user have the option to override these paths

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+        "image-text-to-text",
+        "--trust-remote-code",
+        str(output_dir),
+    ]
+
+    print("[model] exporting:", " ".join(command))
+    subprocess.run(command, check=True)
+
+    if not any(output_dir.glob("*.xml")):
+        raise RuntimeError("OpenVINO export failed, no XML files found")
```
walidbarakat: let's be careful with raising runtime errors without catching these errors and gracefully exit the script

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+
+    return parser.parse_args()
+
+
+def main() -> int:
+    """Entry point."""
+    args = parse_args()
+
+    video_path = ensure_video(args.video)
+    model_path = ensure_model(args.model)
```
walidbarakat: let's use more explicit name for these functions. the word "ensure" can hold different meaning at first glance.

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,76 @@
+# VLM Alerts
+
+This sample demonstrates how to download a Vision-Language Model (VLM) from Hugging Face, export it to OpenVINO IR using `optimum-cli`, and run inference in a DL Streamer pipeline.
```
tjanczak: I'd add more specific information about edge use case... please mention 'alerts' to monitor events/incidents which are hard to identify using traditional computer vision. We can mention 'police car' specifically saying detection of a police car would require dedicated training for traditional computer vision models, while VLM foundational models can detect such events without specialized training.

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## How It Works
+
+The script performs three main steps:
+
+STEP 1 — Prepare input video  
+If a local file is provided, it is used directly.  
+If a URL is provided, the video is downloaded automatically into the `videos/` directory.
+
+STEP 2 — Prepare VLM model
```
tjanczak: Let's be more specific - the model is downloaded from HuggingFace and converted to OpenVINO IR.

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Exported artifacts are stored under:
+
+    models/<ModelName>
+
+STEP 3 — Build and run the pipeline  
+
+The GStreamer pipeline includes:
+
+- gvagenai for VLM inference
```
tjanczak: I'd list elements in the pipeline order... we can also add a graph

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,76 @@
...
+    results/<model>-<video>.jsonl
+
+Annotated video:
+
+    results/<model>-<video>.mp4
+
+## Notes
+
+- Each video and model are downloaded and exported once.
+- Different VLMs can be downloaded. Suggested: OpenGVLab/InternVL3_5-2B, openbmb/MiniCPM-V-4_5, Qwen/Qwen2.5-VL-3B-Instruct.
```
tjanczak: Let's add the list of models to script itself, it should display this information when no arguments are provided.

samples/gstreamer/python/vlm_alerts/requirements.txt
```diff
@@ -0,0 +1,10 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
+PyGObject==3.50.0
+torch==2.9.0+cpu
+transformers==4.57.6
+optimum-intel==1.27.0
+huggingface_hub==0.36.1
+einops
```
tjanczak: do we need specific versions? let's just use the version reported by pip freeze.

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
```diff
@@ -0,0 +1,238 @@
...
+    with os.fdopen(fd, "w") as file:
+        file.write(cfg.question)
+
+    generation_cfg = f"max_new_tokens={cfg.max_tokens}"
+
+    pipeline_str = (
+        f'filesrc location="{cfg.video}" ! '
+        f'decodebin3 ! '
+        f'videoconvertscale ! '
+        f'video/x-raw,format=BGRx,width=1280,height=720 ! '
```
tjanczak, line 165: I'd add width/height as programmable parameters, yet it can be done as next step (another PR)

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
oonyshch: @walidbarakat The function names are fine. Means "ensure that video is being processed"
Same with model
If you have a concrete suggestion, please provide it

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
oonyshch: Here I agree, it would be useful to have arguments in cli
But I am not sure about video-path, video-url 
because then we would also have model-path, model-url
Additionally noticed that prompt would be better than "question", but am still not sure about which arguments  to have here
It also kinda overrides the logic that the dirs for videos and models are created automatically
Do you have an idea? @walidbarakat

walidbarakat: i think having multiple, small and well separated functions is always better. we just need to separate the concerns of the functions i.e downloading and creating dirs shouldn't be in the same function neither one implies calling the other.

 i agree with you that "prompt" is better that "question" and the argument can be just `--prompt`

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
walidbarakat: but this current version of the function doesn't "ensure" the video is being processed in the sense of checking and handling possible error(s) that can stop the processing of the video. it tries to "get" or "find" it on disk or download from remote url. i would suggest `get_video` for example.

samples/gstreamer/python/vlm_alerts/vlm_alerts.py
oonyshch: @walidbarakat Yeah, gotchu, thanks for the good suggestion, changing.

dependencies/gstreamer.cmake
```diff
@@ -39,7 +39,7 @@ ExternalProject_Add(
     	                -Ddevtools=disabled
     	                -Dorc=disabled
     	                -Dgpl=disabled
-                      -Dpython=enabled
+						-Dpython=enabled
```
oonyshch, line 42: I know this is out of scope, just noticed misalignment in formatting.

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,124 @@
...
+- `--video-url` to download automatically
+
+Downloaded videos are cached under `videos/`. 
+
+## Pipeline Architecture
+
+The pipeline is built dynamically in Python using `Gst.parse_launch`.
+
+```mermaid
+graph LR
```
tjanczak, line 53: This is great; we should update other samples to show graphs visually!
I'd suggest to remove some 'auxiliary' elements for better clarity: videoconvertscale, queue, videoconvert... and we could also combine vah264enc/h264parse/mp4mux into 'encode'

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,124 @@
...
+    M --> N[h264parse]
+    N --> O[mp4mux]
+    O --> P[filesink]
+```
+
+## Setup
+
+```bash
+cd samples/gstreamer/python/vlm_alerts
+python3 -m venv .venv
```
tjanczak: I'd suggest to use sample-specific name for virtual environment... if all samples use 'venv' they will start colliding

samples/gstreamer/python/vlm_alerts/README.md
```diff
@@ -0,0 +1,124 @@
...
+source .venv/bin/activate
+pip install -r requirements.txt
```
tjanczak: We have changed other samples that use HuggingFace models to download requirements file from OV repo: 
https://github.com/openvinotoolkit/openvino.genai/raw/refs/heads/releases/2026/0/samples/export-requirements.txt

oonyshch: @tjanczak Thanks, I saw that PR. 
Good idea, but it also imports additional packages, that are pretty heavy, i.e. 
`transformers==4.55.4 # For Whisper`

DL Streamer doesn't use Whisper, right? 
And for my sample I need newer version of transformers.

I would really stay with what I have currently in the PR for the sample instead or installing from two requirements.txt files. What do you think?

samples/gstreamer/python/vlm_alerts/README.md
oonyshch: sure, commiting

samples/gstreamer/python/vlm_alerts/README.md
oonyshch: @tjanczak Thanks, I agree, let's do that. Following up on suggestion, commiting

samples/gstreamer/python/vlm_alerts/README.md
tjanczak: DLStreamer does support whisper in gvaaudiotranscript element... you are right OV file is a 'superset' of what this sample needs.
However, I'm also trying to avoid us managing separate requirements files for each sample, as we can start using conflicting sets. How long does it take to install (excessive) packages?

walidbarakat (approved):
looks good, if this script will be integrated in any automated pipeline or we need to showcase our way of error resiliency we can add a robust error and exception handling functionality. otherwise no need.

PR #638, opened 2026-02-24
[POST-PROC][YOLOv26 OBB] add blob parsing function to handle obb dimension

walidbarakat:
### Description

YOLOv26-OBB outputs center-based coordinates, YOLOv10 blob parsing function was interpreting these as corner-based coordinates.

### Checklist:


src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/yolo_v26.cpp
```diff
@@ -16,6 +16,54 @@
 
 using namespace post_processing;
 
+void YOLOv26ObbConverter::parseOutputBlob(const float *data, const std::vector<size_t> &dims,
```
tjanczak: Is this function different from: YOLOv10Converter::parseOutputBlob?
If same/similar, can we add changes to YOLOv10Converter instead?

walidbarakat: done

tjanczak (approved):

PR #648, opened 2026-02-26
Fix:  threshold parameter in gvadetect does not work with PDD model

mholowni:
### Description

If model is providing confidence threshold, user provided threshold does not work. This PR is allowing user to provide its own detection threshold and overwrite model provided one.

N/A

### How Has This Been Tested?

This was tested with different thresholds provided as gvadetect parameter and checked if there weren't any detections with confidence lower than provided (using PDD model)

### Checklist:


src/monolithic/gst/inference_elements/gvadetect/gstgvadetect.h
```diff
@@ -22,6 +22,7 @@ G_BEGIN_DECLS
 typedef struct _GstGvaDetect {
     GvaBaseInference base_inference;
     double threshold;
+    gboolean threshold_explicitly_set;
```
tjanczak, line 25: why do we need it? why default value does not work?

src/monolithic/gst/inference_elements/common/post_processor.cpp
```diff
@@ -201,6 +201,7 @@ PostProcessor::PostProcessor(InferenceImpl *inference_impl, GvaBaseInference *ba
         initializer.converter_type = ConverterType::TO_ROI;
         GstGvaDetect *gva_detect = reinterpret_cast<GstGvaDetect *>(base_inference);
         initializer.threshold = gva_detect->threshold;
+        initializer.threshold_explicitly_set = gva_detect->threshold_explicitly_set;
```
tjanczak, line 204: why do we need extra argument if threshold was just set in previous line?

src/monolithic/gst/inference_elements/gvadetect/gstgvadetect.h
mholowni: In some models - in this case pallet defect detection, confidence_threshold is defined in model's xml file, so user can't set custom threshold via gvadetect parameter

dmichalo:
Is bigger image size expected?
image

PR #663, opened 2026-03-04
[INFERENCE][YOLO26] fix YOLO26 model inference on GPU FP16/FP32

walidbarakat:
### Description

this commit is a workaround to fix YOLO26 model inference operation on GPU device with precision FP16 or FP32.
the bug was root caused in GPU plugin where the FP16 kernel path bug with Swish activation function

### Checklist:


src/monolithic/inference_backend/image_inference/openvino/openvino_image_inference.cpp
```diff
@@ -1066,10 +1066,40 @@ class OpenVinoNewApiImpl {
...
+            }
+        }
+    }
+
     // Loads network to the device
     void load_network(const ConfigHelper &config) {
         assert(!_compiled_model);
         ov::AnyMap ov_params = config.inference_cfg();
+
+        apply_workaround(config, ov_params);
```
tjanczak: I'm not sure we should degrade precision under the hood... it may be better just generate an error message (inside model_api_converters when we detect an int8 yolov26 and ask users to deliver fp16 version).
I mean int8 model version is effectively not supported, so let's not pretend we support it by falling back to fp32 precision.

tjanczak: Ok, I've read it more thoroughly and I get it as:
fp32 model runs fine (is it the case, I believe OpenVINO tries to run fp32 models as fp16 internal precision)
fp16 model has functional issues (again is it all kind of yolov26 or specific one?)
in8 model runs fine

is it correct?

tbujewsk: This change concerns only FP16. And we are generating warning GVA_WARNING.
But it may be better to replace GVA_WARNING with g_warning as we are doing it in get_model_info_preproc(...) about need of model re-quantization .

walidbarakat: it seems the issue is with YOLO26-OBB FP16 and FP32 only. as i discussed with @tjanczak a runtime error will be thrown if the user wanted to use this model with the relevant quantization as not supported

dmichalo (approved):

PR #716, opened 2026-03-23
Add example pipeline with custom frame selection logic for VLM queries

tjanczak:
### Description

Add a new Python sample which combines CV object detection (yolo26 by default) with GenAI VLM (MiniCPM v4.5 by default). CV path detects object categories - like bottle or apple, and VLM path does detailed item classification (green vs. red apple). The sample demonstrates how CV+VLM pipelines can be constructed with custom logic to select frames eligible for VLM processing. 

Fixes # (issue)

None.

### How Has This Been Tested?

Validated locally on development machine. 

### Checklist:


samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
```diff
@@ -0,0 +1,197 @@
...
+# Copyright (C) 2026 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+# ==============================================================================
+
+"""
+Custom GStreamer element for VLM self-checkout frame selection.
+"""
+
+from dataclasses import dataclass, field
```
oonyshch:
```suggestion
from dataclasses import dataclass
```

samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
```diff
@@ -0,0 +1,197 @@
...
+            1) Select frames with at least one object detected
+            2) Select frames with object tracked for > threshold milliseconds
+            3) Set gvagenai prompt to classify which item from the inventory is visible in the frame"""
+        
+        _, state, _ = self.get_state(0)
+        if state != Gst.State.PLAYING:
+            return Gst.FlowReturn.OK
+        
+        # Convert frame timestamp to milliseconds
+        self._framecount += 1
```
oonyshch: `self._framecount` is incremented
but i don't think is read or exposed as a property

samples/gstreamer/python/vlm_self_checkout/requirements.txt
```diff
@@ -0,0 +1,3 @@
+PyGObject==3.50.0
+--extra-index-url https://download.pytorch.org/whl/cpu
+ultralytics==8.4.7
```
oonyshch: some packages here could be unnecesary, we are for sure importing `PyGObject` when building DLS from sources, please check

oonyshch: also no empty line

src/monolithic/gst/elements/gvagenai/gstgvagenai.cpp
```diff
@@ -367,6 +372,18 @@ static GstFlowReturn gst_gvagenai_transform_ip(GstBaseTransform *base, GstBuffer
         return GST_FLOW_ERROR;
     }
 
+    GST_OBJECT_LOCK(gvagenai);
+    if (gvagenai->prompt_changed) {
+        if (!load_effective_prompt(gvagenai)) {
```
oonyshch, line 377: load_effective_prompt is called while GST_OBJECT_LOCK is held. GST_ELEMENT_ERROR inside it will try to re-acquire the same lock → deadlock. we could snapshot the flag, unlock first, then reload:

```cpp
GST_OBJECT_LOCK(gvagenai);
gboolean need = gvagenai->prompt_changed;
if (need) gvagenai->prompt_changed = FALSE;
GST_OBJECT_UNLOCK(gvagenai);
if (need && !load_effective_prompt(gvagenai)) {
    GST_ELEMENT_ERROR(gvagenai, RESOURCE, FAILED,
        ("Failed to load effective prompt"),
        ("Could not load or validate prompt configuration"));
    return GST_FLOW_ERROR;
}
```

samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
```diff
@@ -0,0 +1,197 @@
...
+            print(f"[gvaframeselection] ERROR: cannot read {label} file '{path}': {e}")
+            return []
+
+    def _resolve_genai_element(self):
+        """Look up the gvagenai element by name from the parent pipeline."""
+        if self._genai_element is not None:
+            return self._genai_element
+        if not self._genai_name:
+            return None
+        pipeline = self.get_parent()
```
oonyshch, line 118: get_parent() returns the immediate parent bin, not the top-level pipeline. if gvagenai lives in a sibling or parent bin, get_by_name() won't find it and prompt updates could stop working. pls consider walking up until reaching a Gst.Pipeline:

```python
p = self.get_parent()
while p and not isinstance(p, Gst.Pipeline):
    p = p.get_parent()
pipeline = p
```

samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
tjanczak: fixed

samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
tjanczak: removed, it was used for debugging but no longer needed

samples/gstreamer/python/vlm_self_checkout/requirements.txt
tjanczak: Installing Python dependencies is an optional step in our setup instruction: https://docs.openedgeplatform.intel.com/dev/edge-ai-libraries/dlstreamer/get_started/install/install_guide_ubuntu.html#optional-step-4-python-dependencies

PyGObjects is required for GStreamer python bindings to work...

src/monolithic/gst/elements/gvagenai/gstgvagenai.cpp
tjanczak: I'm not sure if GST_ELEMENT_ERROR will try to acquire same lock... I believe it resolves to a simple log function: 
https://github.com/GStreamer/gstreamer/blob/390b6dc5a808dbb7c02f4c4565f9244379e263c2/subprojects/gstreamer/gst/gstinfo.h#L748

If all GST logging functions would acquire object log, then it would be hard to use them without deadlocks. Can you please share where GST_ELEMENT_ERROR acquires the object lock?

samples/gstreamer/python/vlm_self_checkout/plugins/python/gvaFrameSelection.py
tjanczak: Right, yet in this example we do not have gvagenai embedded in a bin element, and we don't claim gvaFrameSelection as a generic element to be reused outside of this sample. 
This is just an example demonstrating how one can create a custom 'chunking logic' prior to gvagenai element...

samples/gstreamer/python/vlm_self_checkout/requirements.txt
oonyshch: yeah, we have recently discussed removal of the `[Optional]` there
esp. it says below, in that link:
"
  **required** to run Deep Learning Streamer python elements or samples are not installed by default.
 and then there is a link to install these required packages, among which we have PyGObject
 "
 
 this is minor, just thinking we could either rephrase documentation, revise packages among DLS's requirements.txt
 ...or all of the above 😊

src/monolithic/gst/elements/gvagenai/gstgvagenai.cpp
oonyshch: the lock is not acquired in the logging path. you're right about that. 
but `GST_ELEMENT_ERROR` doesn't only log 
it also calls `gst_element_message_full_with_details()` ([gstelement.c#L2213](https://github.com/GStreamer/gstreamer/blob/main/subprojects/gstreamer/gst/gstelement.c#L2213)), which internally calls `gst_object_get_path_string()` ([gstelement.c#L2246](https://github.com/GStreamer/gstreamer/blob/main/subprojects/gstreamer/gst/gstelement.c#L2246)). and that function is documented as following:
"
Grabs and releases the GstObject's LOCK for all objects in the hierarchy
" 
and calls `gst_object_get_parent()` in a loop ([gstobject.c#L1016-L1040](https://github.com/GStreamer/gstreamer/blob/main/subprojects/gstreamer/gst/gstobject.c#L1016)).
so imo since `GST_OBJECT_LOCK` is a non-recursive mutex, calling it while already holding the lock will deadlock.

tjanczak: ok, I'll fix it in subsequent commit

PR #718, opened 2026-03-24
Enable gvametapublishkafka gvametapublishmqtt on Windows

yunowo:
### Description

Enable gvametapublishkafka and gvametapublishmqtt on Windows
Also add a vcpkg triplet file and build static libraries to reduce DLL count. (dynamic is default)

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
```diff
@@ -26,12 +31,22 @@ GST_DEBUG_CATEGORY_STATIC(gva_meta_publish_mqtt_debug_category);
 
 namespace {
 std::string generate_client_id() {
+#ifdef _WIN32
+    UUID winuuid;
+    UuidCreate(&winuuid);
+    RPC_CSTR szUuid = NULL;
```
oonyshch:
```suggestion
    RPC_CSTR szUuid = nullptr;
```

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
```diff
@@ -26,12 +31,22 @@ GST_DEBUG_CATEGORY_STATIC(gva_meta_publish_mqtt_debug_category);
 
 namespace {
 std::string generate_client_id() {
+#ifdef _WIN32
```
oonyshch: #ifdef inside function body is from c
could we please extract platform implementations into separate .cpp files compiled conditionally, or wrap them in inline helpers declared in a header

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
```diff
@@ -26,12 +31,22 @@ GST_DEBUG_CATEGORY_STATIC(gva_meta_publish_mqtt_debug_category);
...
+    UUID winuuid;
+    UuidCreate(&winuuid);
```
oonyshch: no error handling for `UuidCreate`
modern c++ expects raii or explicit error propagation 
please consider a [[nodiscard]]-aware wrapper.

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
```diff
@@ -26,12 +31,22 @@ GST_DEBUG_CATEGORY_STATIC(gva_meta_publish_mqtt_debug_category);
...
+    RPC_CSTR szUuid = nullptr;
+    UuidToStringA(&winuuid, &szUuid);
+    std::string result(reinterpret_cast<char *>(szUuid));
```
oonyshch: raw c-style reinterpret_cast from unsigned char* to char* masks the type mismatch. prefer std::bit_cast<const char*> (c++20) or at least static_cast through void*, and add a static_assert on sizes.

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
```diff
@@ -26,12 +31,22 @@ GST_DEBUG_CATEGORY_STATIC(gva_meta_publish_mqtt_debug_category);
...
+    RpcStringFreeA(&szUuid);
+    return result;
+#else
     uuid_t binuuid;
     uuid_generate_random(binuuid);
     char uuid[37];
```
oonyshch: magic number 37 for uuid string length with only a comment below to explain what is there
please use a named constexpr

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
walidbarakat: @oonyshch  `std::bit_cast` are for the value conversion with keeping the bit pattern. i.e from `float` to `uint32_t` while preserving the bit pattern. but `std::reinterpret_cast` is definitely for pointer type conversion.

oonyshch: @walidbarakat you are right, my bad. maybe then consider wrapping it into a named inline helper, like:
`rpc_cstr_to_string` that isolates the cast

yunowo: Switched to wchar and no more casts

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
yunowo: Done with a header. Too short for a separate cpp file

src/monolithic/gst/elements/gvametapublish/mqtt/gvametapublishmqtt.cpp
yunowo: Done with return code check. 
This is a quite old Win32 API. For modern C++ we might have to use modern third party libs

PR #719, opened 2026-03-24
Affinity CPU thread affinity parameter to gvadetect/gvaclassify

tjanczak:
### Description

Add new parameter in gvadetect/gvaclassify elements - set CPU affinity mask for stable performance results. 
In addition if no mask is set by a user (either through new parameter or process-wide taskset) and Intel Core Ultra 3xxH is detected, then set affinity mask to p-cores for inference elements. 

Fixes # (issue)

No specific bug filed, but performance results are fluctuating if core pinning is not used with 3xxH series. 

No new dependencies.

### How Has This Been Tested?

Tested locally, to be verified in CI.

### Checklist:


src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
+            }
+        }
+
+    } catch (const std::exception &e) {
+        GST_ELEMENT_ERROR(base_inference, RESOURCE, SETTINGS, ("Invalid core-pinning format"),
+                          ("Failed to parse core-pinning property: %s", e.what()));
+    }
+
+    // if string parsed without errrors, update core pinning
+    base_inference->core_pinning_mask = core_mask;
```
oonyshch, line 776: This line is outside the `try/catch` block. When parsing fails, `core_mask` remains 0 (from line 631 initialization) and gets assigned to `core_pinning_mask`, which silently disables all cores and overwrites any previously valid configuration.

Consider moving the assignment inside the `try` block after successful parsing to preserve the existing value on error:

```suggestion
    } catch (const std::exception &e) {
        GST_ELEMENT_ERROR(base_inference, RESOURCE, SETTINGS, ("Invalid core-pinning format"),
                         ("Failed to parse core-pinning property: %s", e.what()));
        return; // Preserve existing core_pinning_mask on error
    }

    // If string parsed without errors, update core pinning
    base_inference->core_pinning_mask = core_mask;
```

oonyshch (changes requested):

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
+            part.erase(part.find_last_not_of(" \t") + 1);
+
+            if (part.find('-') != std::string::npos) {
+                // Parse range like "1-5"
+                size_t dash_pos = part.find('-');
+                int start = std::stoi(part.substr(0, dash_pos));
+                int end = std::stoi(part.substr(dash_pos + 1));
+
+                // Set bits in the range
+                for (int i = start; i <= end; i++) {
```
oonyshch: Missing bounds validation causes undefined behavior when core index >= 64. Shifting a 64-bit value by >= 64 bits is UB per C++ standard. Consider adding validation:

```suggestion
                // Set bits in the range
                for (int i = start; i <= end; i++) {
                    if (i >= 0 && i < 64) {
                        core_mask |= (1ULL << i);
                    } else {
                        GST_WARNING_OBJECT(base_inference, "Core index %d out of range [0-63], skipping", i);
                    }
                }
```

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
+                    core_mask |= (1ULL << i);
+                }
+            } else {
+                // Set single bit
+                core_mask |= (1ULL << std::stoi(part));
```
oonyshch: Same issue as the range loop: missing bounds check before bit shift. Consider adding validation:

```suggestion
                // Set single bit
                int core_id = std::stoi(part);
                if (core_id >= 0 && core_id < 64) {
                    core_mask |= (1ULL << core_id);
                } else {
                    GST_WARNING_OBJECT(base_inference, "Core index %d out of range [0-63], skipping", core_id);
                }
```

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -80,6 +80,8 @@
 
 #define DEFAULT_SHARE_VADISPLAY_CTX TRUE
 
+#define DEFAULT_CORE_PINNING nullptr
+
```
oonyshch: The magic number 64 appears throughout the code. Consider defining it as a constant for maintainability:

```suggestion
#define DEFAULT_CORE_PINNING nullptr
#define MAX_CPU_CORES 64
```

Then use `MAX_CPU_CORES` in all the bounds checks and getter loop instead of hardcoded 64.

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -837,6 +892,19 @@ void gva_base_inference_get_property(GObject *object, guint property_id, GValue
     case PROP_SCHEDULING_POLICY:
         g_value_set_string(value, base_inference->scheduling_policy);
         break;
+    case PROP_CORE_PINNING: {
```
oonyshch, line 1081: The getter outputs individual cores ("0,1,2,3") while the setter accepts range notation ("0-3"). This asymmetry means:
- Setting "0-3" then getting returns "0,1,2,3" 
- Copy-pasting the getter output works, but loses the compact representation

Consider implementing range compaction in the getter to match the setter's format. For example:
- Consecutive cores 0,1,2,3 → "0-3"
- Non-consecutive 0,2,5,6,7 → "0,2,5-7"

This makes the property symmetric and more user-friendly for large core counts.

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
+    try {
+        // Split by comma to get individual ranges or numbers
+        std::string str(range_str);
+        std::istringstream iss(str);
+        std::string part;
+
+        // parse input string and set bits in core_pinning accordingly
+        while (std::getline(iss, part, ',')) {
+            // Trim whitespace
+            part.erase(0, part.find_first_not_of(" \t"));
```
oonyshch: Whitespace trimming only handles spaces and tabs. Consider using a more robust approach that handles all whitespace characters (including newlines, carriage returns, etc.):

```suggestion
            // Trim whitespace
            auto is_space = [](unsigned char c) { return std::isspace(c); };
            part.erase(0, std::find_if_not(part.begin(), part.end(), is_space) - part.begin());
            part.erase(std::find_if_not(part.rbegin(), part.rend(), is_space).base(), part.end());
```

This handles all standard whitespace characters and is more maintainable.

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
 }
 
+// Convert range of integer IDs to bitset representing cores for pinning.
+// Example input: "1-5, 8, 10-12"
+void gva_base_inference_set_core_pinning(GvaBaseInference *base_inference, const gchar *range_str) {
+    guint64 core_mask = 0;
+
+    try {
+        // Split by comma to get individual ranges or numbers
+        std::string str(range_str);
```
oonyshch: This creates a full copy of the input string, then `std::istringstream` on line 636 copies it again. For a read-only parameter, this can be wasteful.

Since this is C++17 code (using `std::from_chars` would be better anyway), consider using `std::string_view` for the parameter:

```cpp
void gva_base_inference_set_core_pinning(GvaBaseInference *base_inference, 
                                          const gchar *range_str) {
    std::string_view sv(range_str);
    // Then parse sv directly without copying
}
```

Or better yet, pass through to a helper function that takes `std::string_view` to avoid C string → string_view conversion overhead in the common path. 📎

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
...
+        // parse input string and set bits in core_pinning accordingly
+        while (std::getline(iss, part, ',')) {
+            // Trim whitespace
+            part.erase(0, part.find_first_not_of(" \t"));
+            part.erase(part.find_last_not_of(" \t") + 1);
+
+            if (part.find('-') != std::string::npos) {
+                // Parse range like "1-5"
+                size_t dash_pos = part.find('-');
+                int start = std::stoi(part.substr(0, dash_pos));
```
oonyshch: `std::stoi` throws exceptions on invalid input, which is expensive for control flow. For performance-critical parsing, use C++17's `std::from_chars` which is:
- Exception-free (returns error code)
- Significantly faster (~10x in benchmarks)
- Doesn't allocate or use locale

```cpp
#include <charconv>

int start, end;
auto [ptr1, ec1] = std::from_chars(part.data(), part.data() + dash_pos, start);
auto [ptr2, ec2] = std::from_chars(part.data() + dash_pos + 1, part.data() + part.size(), end);
if (ec1 == std::errc{} && ec2 == std::errc{}) {
    // Valid parse
} else {
    // Handle error without exception overhead
}
```

Applies to lines 648, 649, and 657.

src/monolithic/gst/inference_elements/base/inference_impl.cpp
```diff
@@ -977,6 +986,58 @@ bool InferenceImpl::IsRoiSizeValid(const GstAnalyticsODMtd roi_meta) {
...
+ * Pins current thread to the CPU core set as specified by affinity mask.
+ */
+void InferenceImpl::SetAffinityMask(uint64_t mask) {
+    GVA_INFO("Setting CPU affinity mask to 0x%lx\n", mask);
+#ifndef _WIN32
+    cpu_set_t cpuset;
+    CPU_ZERO(&cpuset);                             // Initialize to zero
+    int num_cores = sysconf(_SC_NPROCESSORS_ONLN); // Get number of available CPU cores
+    for (int core_id = 0; core_id < num_cores; ++core_id) {
+        if (mask & (1ULL << core_id)) {
```
oonyshch: Here mask can have bits set beyond `num_cores`

The input `mask` parameter (64-bit) can have bits set for cores that don't exist on the system. For example:
- System has 16 cores (`num_cores = 16`)
- User passes mask `0xFFFFFFFFFFFFFFFF` (all 64 bits set)
- Loop iterates 16 times, but bits 16-63 in mask are never validated
- `CPU_SET` gets called only for existing cores, BUT the mask is accepted silently

This creates inconsistency: the user thinks they pinned to cores 0-63, but actually only pinned to 0-15.

Please consider validating mask against actual core count:
```cpp
// After line 1016
if (mask >= (1ULL << num_cores)) {
    GVA_WARNING("Affinity mask 0x%lx has bits set beyond available cores (%d), truncating\n", mask, num_cores);
    mask &= ((1ULL << num_cores) - 1);  // Clear invalid bits
}
```

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -615,6 +625,48 @@ void gva_base_inference_set_labels(GvaBaseInference *base_inference, const gchar
     }
 }
 
+// Convert range of integer IDs to bitset representing cores for pinning.
+// Example input: "1-5, 8, 10-12"
+void gva_base_inference_set_core_pinning(GvaBaseInference *base_inference, const gchar *range_str) {
+    guint64 core_mask = 0;
```
mholowni: What about a possibility that we have a CPU with more than 64 cores?

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
jmotow: Good observation on shift safety! But I believe on Linux CPU_SET is safe here — it operates on a 1024-bit cpu_set_t and internally uses modulo 64 for the shift amount, so the shift is always within bounds regardless of the core index. The existing i < num_cores check is sufficient to stay within [0, CPU_SETSIZE).

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
```diff
@@ -490,6 +498,77 @@ void gva_base_inference_cleanup(GvaBaseInference *base_inference) {
...
+    memset(&base_inference->core_pinning_mask, 0, sizeof(WinCorePinningMask));
+    WORD num_groups = GetActiveProcessorGroupCount();
+    base_inference->core_pinning_mask.num_groups = num_groups;
+
+    // Default: set all cores in all groups
+    int total_cores = 0;
+    for (WORD g = 0; g < num_groups; ++g) {
+        DWORD count = GetActiveProcessorCount(g);
+        total_cores += count;
+        base_inference->core_pinning_mask.group_mask[g] = ((KAFFINITY)1 << count) - 1;
```
jmotow, line 555: I believe when count == 64, the _**shift (KAFFINITY)1 << count**_ is undefined behavior. Maybe it would be good to check the size first, something like that(?):

**_if (count >= sizeof(KAFFINITY) * 8)
        base_inference->core_pinning_mask.group_mask[g] = ~(KAFFINITY)0;
else
    base_inference->core_pinning_mask.group_mask[g] = ((KAFFINITY)1 << count) - 1;_**

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
tbujewsk: good point, to be addressed as separate PR

src/monolithic/gst/inference_elements/base/gva_base_inference.cpp
tbujewsk: good point, also many other params of different elements should be considered; to be handle via separate PR

mholowni (dismissed):

oonyshch (dismissed):

xwu2intel:
Do not hard-code why core pinning in DLS, which will yield counter-intuitive results.

mholowni (approved):

oonyshch (approved):

PR #723, opened 2026-03-25
Update SPECS files

mholowni:
### Description

Update versions in specs

### Checklist:

### DIFF
--- a/SPECS/intel-dlstreamer/intel-dlstreamer.spec
+++ b/SPECS/intel-dlstreamer/intel-dlstreamer.spec
@@ -20,7 +20,7 @@ BuildRequires:  libva-devel libva-intel-media-driver
 BuildRequires:  python3-devel python3-pip
 BuildRequires:  pkgconfig patchelf
 BuildRequires:  opencv-devel >= 4.12.0
-BuildRequires:  gstreamer-devel >= 1.26.1
+BuildRequires:  gstreamer-devel >= 1.26.6
 BuildRequires:  paho-mqtt-c-devel >= 1.3.4
 BuildRequires:  librdkafka-devel
 


SPECS/intel-dlstreamer/intel-dlstreamer.spec
```diff
@@ -139,6 +139,8 @@ rm -rf %{buildroot}
 /opt/intel/dlstreamer/lib/pkgconfig/*
 
 %changelog
+* Tue Mar 24 2025 DL Streamer Team <dlstreamer@intel.com> - 2026.0.0-1
```
oonyshch:
```suggestion
* Wed Mar 25 2026 DL Streamer Team <dlstreamer@intel.com> - 2026.0.0-1
```

SPECS/intel-dlstreamer/intel-dlstreamer.spec
oonyshch:
```suggestion
BuildRequires:  gstreamer-devel >= 1.26.6
```
should this also be 1.26.6?

SPECS/intel-dlstreamer/intel-dlstreamer.spec
```diff
@@ -27,7 +27,7 @@ BuildRequires:  librdkafka-devel
 # Runtime dependencies
 Requires:       paho-mqtt-c-devel >= 1.3.4
```
oonyshch, line 28: also here we are potentially mixing base-package vs devel packages:
the runtime Requires maybe should use `paho-mqtt-c` (base package), not `paho-mqtt-c-devel`., as above in BuildRequires ?

as the `-devel` variant is only needed at build time (headers, .so...), maybe pulling it in as a runtime dependency is not needed

overall i went on a tangent looking for how exactly this one should look, and not sure

mholowni: To be honest I am not sure, SPECs are not created by me, I am only updating it

oonyshch: neither am I, let's leave as is for now then, as we are updating a different thing

PR #725, opened 2026-03-26
Optimizer pipelines will now be validated based on the counted detections.

pbartosik:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


scripts/optimizer/optimizer.py
```diff
@@ -166,9 +183,12 @@ def _optimize_pipeline(self, starting_pipeline, starting_fps, start_time, stream
                     pipelines.append(pipeline)
 
                 try:
-                    fps = sample_pipeline(pipelines, self._sample_duration)
+                    fps, detections = sample_pipeline(pipelines, self._sample_duration)
+
+                    if detections / starting_detections < self._detections_error_margin:
+                        logger.debug("Pipline reporting detections under error margin, skipping")
```
oonyshch:
```suggestion
                        logger.debug("Pipeline reporting detections under error margin, skipping")
```

oonyshch: this debug log seems to not actually skip the pipeline since there's no `continue` statement after it. 
the actual skipping happens on line 191 with the `and` condition. maybe we should either:

1. add `continue` after this log to actually skip, or
2. remove this `if` block entirely since line 191 already handles it

otherwise that only logs but never skips

scripts/optimizer/optimizer.py
```diff
@@ -270,10 +291,3 @@ def process_bus(bus):
             logger.error("Other message: %s", str(message))
         message = bus.pop()
 
-##################################### Compatibility ###############################################
-
-def get_optimized_pipeline(pipeline, search_duration = 300, sample_duration = 10):
```
oonyshch, line 275: thought this was here for backward compatibility for any external code using `get_optimized_pipeline()`
why are we removing it? maybe we should change it to counting detections

scripts/optimizer/optimizer.py
pbartosik: good catch, done

scripts/optimizer/optimizer.py
pbartosik: This was indeed here for backwards compatibility, specifically to not break VIPPET code. They have since adapted their code, so i opted to remove it so that new users wouldn't attempt to use this function since deprecation warnings are a little obtuse in python.

scripts/optimizer/optimizer.py
```diff
@@ -166,7 +183,12 @@ def _optimize_pipeline(self, starting_pipeline, starting_fps, start_time, stream
                     pipelines.append(pipeline)
 
                 try:
-                    fps = sample_pipeline(pipelines, self._sample_duration)
+                    fps, detections = sample_pipeline(pipelines, self._sample_duration)
+
+                    if ((detections == 0 and detections != starting_detections)
```
oonyshch: potential division by zero when `starting_detections == 0` ?
maybe :

```python
if starting_detections == 0:
    # skip only if we still have zero detections
    if detections == 0:
        logger.debug("Pipeline reporting detections under error margin, skipping")
        continue
elif detections / starting_detections < self._detections_error_threshold:
    logger.debug("Pipeline reporting detections under error margin, skipping")
    continue
```

scripts/optimizer/optimizer.py
```diff
@@ -78,13 +82,12 @@ def optimize_for_fps(self, pipeline):
...
             logger.info("Measuring performance of the original pipeline...")
-            fps = sample_pipeline([pipeline], self._sample_duration)
+            initial_fps, initial_detections = sample_pipeline([pipeline], self._sample_duration)
         except Exception as e:
             logger.error("Pipeline failed to start, unable to measure fps: %s", e)
             raise RuntimeError("Provided pipeline is not valid") from e
 
-        logger.info("FPS: %.2f", fps)
-        initial_fps = fps
+        logger.info("FPS: %.2f", initial_fps)
```
p-zak, line 90: It seems that logging also `initial_detections` here would be useful for the user.

scripts/optimizer/optimizer.py
```diff
@@ -122,6 +125,16 @@ def optimize_for_streams(self, initial_pipeline):
...
 
+        # Measure the performance of the original pipeline
+        try:
+            logger.info("Measuring performance of the original pipeline...")
+            initial_fps, initial_detections = sample_pipeline([initial_pipeline], self._sample_duration)
+        except Exception as e:
+            logger.error("Pipeline failed to start, unable to measure fps: %s", e)
+            raise RuntimeError("Provided pipeline is not valid") from e
+
+        logger.info("FPS: %.2f", initial_fps)
```
p-zak, line 136: It seems that logging also `initial_detections` here would be useful for the user.

scripts/optimizer/optimizer.py
```diff
@@ -250,8 +276,9 @@ def sample_pipeline(pipelines, sample_duration):
 
     del pipeline
     fps = fps_counter.get_property("avg-fps")
+    detections = fps_counter.get_property("detections")
     logger.debug("Sampled fps: %.2f", fps)
-    return fps
+    return fps, detections
```
p-zak, line 281: It seems that logging also `detections` here would be useful for the user.

scripts/optimizer/optimizer.py
```diff
@@ -166,7 +183,16 @@ def _optimize_pipeline(self, starting_pipeline, starting_fps, start_time, stream
...
+                    fps, detections = sample_pipeline(pipelines, self._sample_duration)
+
+                    if starting_detections == 0:
+                        # skip only if we still have zero detections
+                        if detections == 0:
+                            logger.debug("Pipeline reporting detections under error margin, skipping")
+                            continue
+                    elif detections / starting_detections < self._detections_error_threshold:
+                        logger.debug("Pipeline reporting detections under error margin, skipping")
+                        continue
```
p-zak, line 195: Do we expect the pipeline to be skipped quite often, or do we treat this as an incidental situation?
If it's the latter, then maybe the user should be informed about it at the standard logging level (`info`)?
Or perhaps it should even be a `warning`?

scripts/optimizer/optimizer.py
pbartosik: Generally we don't really want to expose the detections information to the user during normal operations. It's not a metric that explains anything about the optimization progress and its presence could end up being confusing.

scripts/optimizer/optimizer.py
pbartosik: It should (hopefully 😄) be an incidental situation, and no, it wouldn't be a warning since the user hasn't done anything wrong nor is the application in an incorrect state. A potential pipeline returning a significantly reduced amount of detections could only be caused by optimizer's operations and skipping it is just a normal step.

PR #735, opened 2026-03-31
Fix: Add mapping of va memory in gvawatermark

mholowni:
### Description

Add mapping of VAmemory to watermark - to be consistent with d3d11 implementation and proper GPU-CPU memory copy

### Checklist:


src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
```diff
@@ -871,25 +871,27 @@ int Impl::get_num_primitives() const {
 bool Impl::render(GstBuffer *buffer) {
     ITT_TASK(__FUNCTION__);
 
-    // For D3D11 input, map to system memory temporarily for rendering
+    // Skip render if there are no primitives to draw
+    if (prims.empty()) {
```
oonyshch, line 875: this early return now skips gpu memory mapping entirely when no primitives to draw. previously, the buffer would still be mapped/unmapped even with empty prims (the check was inside the d3d11 block). is this intentional? could skipping the map/unmap cycle cause issues if callers expect the buffer to be touched?

src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
```diff
@@ -871,25 +871,27 @@ int Impl::get_num_primitives() const {
...
+    // For GPU memory input, map to system memory temporarily for CPU rendering
     GstMapInfo map_info;
     bool mapped = false;
 
-    if (_mem_type == InferenceBackend::MemoryType::D3D11) {
-        // Map D3D11 buffer to system memory for CPU rendering
+    if (_mem_type == InferenceBackend::MemoryType::D3D11 || _mem_type == InferenceBackend::MemoryType::VAAPI) {
+        // Map GPU buffer to system memory for CPU rendering
         if (!gst_buffer_map(buffer, &map_info, GST_MAP_READWRITE)) {
             return false;
```
oonyshch, line 887: consider adding a gst_error_object or gst_warning_object log here before returning false, to help diagnose va/d3d11 memory mapping failures at runtime.

src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
mholowni: It is intentional, to not map when it is not needed

src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
mholowni: added in 8093cc6b2eec7a6c5697dd9b5ab6031ffe0391a0

dmichalo (changes requested):
Seems like the whole CI is failing

Image

mholowni:
@dmichalo it should be ok now

oonyshch (approved):

dmichalo (approved):

PR #741, opened 2026-04-01
DLStreamer coding agent (draft implementation for validation purposes).

tjanczak:
### Description

Instructions for AI Coding Agents on how to generate new Python sample apps. 
Work in progress - PR submitted for validation purposes on 'clean machines'. 

Fixes # (issue)

No new dependencies.

### How Has This Been Tested?

Experimental feature, work in progress. 

### Checklist:

### DIFF
--- a/.github/workflows/dls-pr-workflow.yaml
+++ b/.github/workflows/dls-pr-workflow.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - 'main'
     paths:
+      - '.github/**'
       - 'cmake/**'
       - 'docker/**'
       - 'include/**'
@@ -16,6 +17,7 @@ on:
       - 'thirdparty/**'
   pull_request:
     paths:
+      - '.github/**'
       - 'cmake/**'
       - 'docker/**'
       - 'include/**'
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -12,7 +12,7 @@ on:
   schedule:
     - cron: '19 4 * * 0'
   push:
-    branches: ["master"]
+    branches: ["main"]
 permissions: read-all
 
 jobs:


.github/skills/dlstreamer-coding-agent/SKILL.md
qianlongding, line 1: In skill.md, do we need to explicitly include Windows-specific requirements? For instance, highlighting that D3D11 is the required path for Windows environments to ensure the AI agent generates valid, platform-compatible pipelines.

qianlongding, line 1: Skill Granularity: Regarding the scope of the skills—should we aim for a single, comprehensive skill that handles all supported sample generation? Or would it be more effective to split them by domain? For example, one skill dedicated to standard Video Analytics applications and another specialized for 3D/LiDAR sample generation.

tjanczak: We should add Windows-specific section or requirements... this is missing now; please add if you could improve it. 
(we will merge this version soon, but feel free to submit additional PRs to add Windows)

tjanczak: I was more thinking about separating skills into:
1) Installation skill (OS-specific)
3) Application development skill (should be mostly common, I'd just add 3D/LIDAR as another flavor of pipelines)
2) Debugging skill (should be mostly common)

.github/skills/dlstreamer-coding-agent/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+| g3dlidarparse | LiDAR parsing pipeline | `samples/gstreamer/gst_launch/g3dlidarparse/` |
+| gvarealsense | RealSense camera capture | `samples/gstreamer/gst_launch/gvarealsense/` |
+| custom_postproc/detect | Custom detection post-processing library | `samples/gstreamer/gst_launch/custom_postproc/detect/` |
+| custom_postproc/classify | Custom classification post-processing library | `samples/gstreamer/gst_launch/custom_postproc/classify/` |
+| face_detection_and_classification_bins | Detection + classification using `processbin`, GPU/CPU VA memory paths | `samples/gstreamer/gst_launch/face_detection_and_classification_bins/` |
+| motion_detect | Motion region detection (`gvamotiondetect`), ROI-restricted inference | `samples/gstreamer/gst_launch/motion_detect/` |
+
+## Procedure
+
+### Quick Recipes
```
oonyshch, line 88: the pattern ids in the table below , like `1 + 4 + 11` require the reader to cross-reference design-patterns.md to understand what they mean.

 so i suggest inlining the names to make the table self-contained, e.g.:
`1 (pipeline core) + 4 (ai inference) + 11 (file output)` 

this also reduces the risk of the agent misreading the recipe if pattern numbering ever shifts.

.github/skills/dlstreamer-coding-agent/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+
+If the user asks for a command-line application, construct a `gst-launch-1.0` pipeline string using the identified DLStreamer elements. Follow established conventions for element properties, caps negotiation, and metadata handling as seen in the reference command line samples.
+
+### Step 3b [Python Application] — Decompose the User Request into Design Patterns
+
+If the user asks for a Python application or wants to add custom logic as new Python elements, decompose the requested pipeline into one or more of the design patterns listed in the [Design Patterns Reference](./references/design-patterns.md). This will guide the structure of the application, including how to construct the pipeline, where to add callbacks, and how to handle models and metadata.
+
+Map the user's description to one or more of these patterns:
+
+| Pattern | When to Apply |
```
oonyshch, line 151: this table lists patterns by name only, while the quick recipes above reference them by number only. an agent reasoning top-to-bottom through this file sees numbers first, then names, with no explicit link between them. adding the pattern numbers as a column here (or as a prefix in the pattern name column) would close that gap.

.github/skills/dlstreamer-coding-agent/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+Note DLStreamer container does not come with render or accel group permissions by default, so you need to add them at runtime using `--group-add` flag and `stat` command to query the correct group ID for your system. For example:
+
+```bash
+docker run -it --rm \
+    -u "$(id -u):$(id -g)" \
+    -v "$(pwd)":/app -w /app \
+    --device /dev/dri \
+    --group-add $(stat -c "%g" /dev/dri/render*) \
+    --device /dev/accel \
+    --group-add $(stat -c "%g" /dev/accel/accel*) \
```
oonyshch, line 205: `stat -c "%g" /dev/accel/accel*` will fail on systems without an npu because the glob expands to nothing. please maybe wrap the `--device /dev/accel` and `--group-add` lines in a conditional, or at minimum add a note that these lines should only be included when npu devices are present on the host.

.github/skills/dlstreamer-coding-agent/assets/python-app-template.py
```diff
@@ -0,0 +1,167 @@
...
+    # Locate models (adjust glob patterns for your models)
+    model_xml = find_model("**/*.xml", "detection")
+
+    # Output dirs
+    Path(args.output_video).parent.mkdir(parents=True, exist_ok=True)
+    Path(args.output_json).parent.mkdir(parents=True, exist_ok=True)
+
+    # GPU fallback
+    device = args.device
+    if device == "GPU" and not os.path.exists("/dev/dri/renderD128"):
```
oonyshch, line 139: the gpu detection relies on `/dev/dri/renderD128` which is the first intel igpu render node on linux. this could manifest as a silent callback to cpu on systems where the render node is renderD129 or higher (like, discrete gpu), and it won't work at all on windows. a more portable approach would be to query openvino directly, like "GPU" in Core().available_devices

as i was typing this, i remembered @pbartosik found a way to do this in an even better way in Optimizer.

.github/skills/dlstreamer-coding-agent/references/coding-conventions.md
```diff
@@ -0,0 +1,155 @@
...
+signal.signal(signal.SIGINT, _sigint_handler)
+```
+
+## GPU Availability Check
+
+Check for GPU availability before constructing the pipeline and fall back to CPU:
+
+```python
+det_dev = args.device
+if det_dev == "GPU" and not os.path.exists("/dev/dri/renderD128"):
```
oonyshch, line 152: same issue as in the `python-app-template`

same issue as in the app template: the recommended gpu check is hardcoded to /dev/dri/renderD128. since this is the canonical convention document that all generated apps will follow, fixing it here would propagate the improvement automatically

.github/skills/dlstreamer-coding-agent/assets/README-template.md
```diff
@@ -0,0 +1,159 @@
+# {{APP_TITLE}}
+
+{{APP_DESCRIPTION}}
+
+![{{APP_TITLE}}]({{APP_IMAGE}})
```
oonyshch, line 5: `{{APP_TITLE}}` is a required-looking line with no guidance on what to do when no screenshot exists yet. a newly generated app won't have one. i suggest adding a comment like `<!-- remove this line if no screenshot is available -->` to prevent the agent from leaving a broken image reference in the final readme

.github/skills/dlstreamer-coding-agent/references/model-preparation.md
```diff
@@ -0,0 +1,298 @@
...
+Source: `samples/gstreamer/python/vlm_alerts/vlm_alerts.py`
+
+Recommended small models for edge: `OpenGVLab/InternVL3_5-2B`, `openbmb/MiniCPM-V-4_5`,
+`Qwen/Qwen2.5-VL-3B-Instruct`, `HuggingFaceTB/SmolVLM2-2.2B-Instruct`.
+
+### 5. PaddlePaddle OCR Models
+
+**When to use:** User asks for OCR (PaddleOCR), or any PaddlePaddle model from HuggingFace.
+
+**CRITICAL:** PaddlePaddle v3+ models use PIR format (`.json` + `.pdiparams`), **not** the
```
oonyshch, line 176: the note that paddleocr v3+ uses pir format `(.json + .pdiparams)` instead of `.pdmodel` is in the export template. could cause failures during model prep. 

please move it to it to the surface explicitly in the paddleocr section of `model-preparation.md` so the agent encounters it during step 2 research, not only when copying the template.

PR #759, opened 2026-04-10
Initial version of dynamic ONVIF camera discovery with DLS pipeline startup

jmotow:
Initial version of dynamic ONVIF camera discovery with DLS pipeline startup


samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_sample.py
```diff
@@ -1,152 +1,93 @@
...
-    prepared_command = ["gst-launch-1.0", "rtspsrc", f"location={camera_rtsp_url}"]
+    engine.init_discovery(cmd_line_params)
 
-    # Safely parse pipeline elements
-    prepared_command.extend(shlex.split(pipeline_elements))
-
-    return prepared_command
+    try:
+        async for _ in engine.discover_cameras_iter():
+            await asyncio.sleep(0)
```
msmiatac: `await asyncio.sleep(0)` is usually not something we want in normal async control flow. If `discover_cameras_iter()` already has real `await` points internally, this extra yield is redundant. If it does not, this line does not make the loop truly async, it only masks the issue. I’d suggest simplifying this to:
```python
try:
    async for _ in engine.discover_cameras_iter():
        pass
finally:
    await engine.release_resources_async()
```
That makes the sample easier to read and keeps the async boundary in the engine, where it belongs.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_discovery_engine.py
```diff
@@ -0,0 +1,691 @@
...
+    Usage::
+
+        async for camera in discover_onvif_cameras_async():
+            print(camera)
+
+    Yields:
+        dict: ``{"hostname": str, "port": int}`` for every discovered camera.
+    """
+    loop = asyncio.get_running_loop()
+    cameras = await loop.run_in_executor(None, lambda: discover_onvif_cameras(verbose))
```
msmiatac: Using `run_in_executor()` here is a valid way to keep the event loop responsive while wrapping blocking discovery logic. However, this still awaits the complete discovery pass before yielding any results, so the function is asynchronous but not incremental/streaming. `run_in_executor()` runs the whole blocking discovery function in a worker thread and only returns once that full pass is done, so callers do not receive cameras as they are found. I’d either update the docstring to describe this as a thread-offloaded wrapper, or switch to an `asyncio.Queue` pattern so the worker thread can publish cameras incrementally. For example:
```python
async def discover_onvif_cameras_async(verbose: bool = False):
    loop = asyncio.get_running_loop()
    queue: asyncio.Queue[dict | None] = asyncio.Queue()

    def worker():
        for camera in discover_onvif_cameras(verbose):
            loop.call_soon_threadsafe(queue.put_nowait, camera)
        loop.call_soon_threadsafe(queue.put_nowait, None)

    threading.Thread(target=worker, daemon=True).start()

    while True:
        item = await queue.get()
        if item is None:
            break
        yield item
```

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_discovery_thread.py
```diff
@@ -0,0 +1,91 @@
...
+        self.definition = definition
+        self.name = name
+        self.pipeline = None
+        self.loop = GLib.MainLoop()
+        self.thread = None
+        self.profile = None
+        self.ip_address: str = ""
+        self._lifecycle_lock = threading.Lock()
+
+    async def start(self):
```
msmiatac: `start()` is declared as `async`, but it does not `await` anything. That makes the API look asynchronous even though the body runs synchronously. This is a confusing pattern because `async def` usually implies cooperative suspension points. If startup is actually synchronous, I’d make this a normal method:
```python
def start(self):
    ...
```
and if we still want to keep the event loop responsive, call it like this:
```python
pipeline_task = asyncio.create_task(asyncio.to_thread(pipeline.start))
```
It means:

1. run pipeline.start() in a worker thread
2. start it in the background right now
3. keep a task handle so the engine can track/cancel/await it later

That makes the design more explicit: blocking work is blocking, and we offload it intentionally.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_discovery_engine.py
```diff
@@ -0,0 +1,691 @@
...
+                    print(f"[ERROR] Failed to stop pipeline for '{camera_id}': {err}")
+                entry.mark_removed()
+                removed_dicts.append(entry.raw_discovery)
+
+        if removed_dicts:
+            print_cameras("Removed cameras", removed_dicts)
+
+        gc.collect()
+
+    def _create_pipelines_for_entry(self, entry: DlsOnvifCameraEntry) -> None:
```
msmiatac: This method is part of the async discovery flow, but it still performs blocking ONVIF calls directly on the event loop thread. That means the top-level API looks async while some of the heaviest work may still block other tasks. A simple improvement would be to isolate the blocking ONVIF interaction in a synchronous helper and offload it with `asyncio.to_thread()`:
```python
def _load_profiles_for_entry_sync(self, entry):
    camera_obj = ONVIFCamera(entry.hostname, entry.port, self.username, self.password)
    return self.camera_profiles(camera_obj)

profiles = await asyncio.to_thread(self._load_profiles_for_entry_sync, entry)
```
That keeps the event loop responsive and also makes the async/sync boundary much clearer for future readers.

In the current form, this blocks the event loop while each camera is being processed, so the discovery loop is async at the top level but still serializes on per-camera setup

oonyshch:
no unit tests are included for the new async discovery engine. given the lifecycle complexity (cameras appearing/disappearing, pipeline start/stop, rollback on failure), consider adding tests for at least: (1) camera discovered triggers pipeline start, (2) camera lost triggers pipeline stop, (3) startup failure triggers `_cleanup_on_failed_start` correctly. mock-based tests would significantly improve confidence here.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_discovery_thread.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _join_thread_or_raise(self, context: str) -> None:
+        """Join worker thread and fail loudly if it doesn't stop in time."""
+
+        if self.thread is None:
+            return
+
+        self.thread.join(timeout=self._THREAD_STOP_TIMEOUT_SEC)
+        if self.thread.is_alive():
+            raise RuntimeError(
```
oonyshch, line 71: potential issue: `_join_thread_or_raise` raises a runtimeerror if the thread doesn't stop within 2 seconds, but the caller (`stop()`) holds `_lifecycle_lock` at that point. if this exception is caught somewhere up the call stack and `stop()` is retried, the lock is still held which could deadlock. consider releasing resources without raising here, or restructuring so the lock is released before surfacing the error.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_config_manager.py
```diff
@@ -0,0 +1,82 @@
...
+    def __del__(self):
+        """Clean up resources held by the configuration manager."""
+        self.cameras.clear()
+        self._config_file_path = None
+
+    def _load_config(self):
+        """Load the configuration from the JSON file into the cameras dict"""
+        try:
+            if self._config_file_path:
+                with open(self._config_file_path, "r", encoding="utf-8") as file:
```
oonyshch, line 33: the config file is opened directly without sanitizing or validating `json_file_path`. if the path is passed from an external source (like cli arg), consider adding a check that the resolved path stays within an expected directory (like `Path(path).resolve().is_relative_to(expected_dir)`) to prevent path traversal.

oonyshch (commented):
overall, good. please take attention before merge:
- two required ci checks: license header scan and pylint. run pylint locally on all new python files and fix issues.
- verify license header format matches existing files, also for CI check
- rebase branch with main.

**additional recommendations:**
- add unit tests for the async discovery lifecycle (camera discovered/lost, pipeline start/stop, failure rollback).
- see inline comments for potential deadlock issue and path traversal security concern.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_config_manager.py
jmotow: Good point as a general security principle, but I don't think it applies here. The json_file_path originates from a CLI argument (--config-file) in dls_onvif_sample.py — the user who runs this script already has full filesystem access. There's no privilege boundary to cross: if someone can pass --config-file=../../etc/passwd, they can just as easily cat /etc/passwd directly. Adding path traversal protection for a local CLI tool with no trust boundary would be over-engineering without a real security benefit.

That said, if this path ever starts coming from an untrusted source (HTTP request, API call, multi-tenant input), adding a resolved-path check would absolutely be warranted.

samples/gstreamer/python/onvif_cameras_discovery/dls_onvif_discovery_thread.py
jmotow: Thanks for flagging this, but there's no deadlock risk here. The lock is acquired via with self._lifecycle_lock:, so Python's context manager protocol guarantees that lock.release() is called on any exit from the block — including unhandled exceptions. If _join_thread_or_raise raises a RuntimeError, the with block's __exit__ fires, the lock is released, and the exception propagates normally. A subsequent retry of stop() would re-acquire the lock without issue.

The caller in stop_all_pipelines() already wraps pipeline.stop() in a try/except, so the exception is caught cleanly with the lock already released.

tbujewsk (dismissed):

tbujewsk (approved):

msmiatac (approved):

PR #769, opened 2026-04-13
Converting gvapython bash samples to Python

jmotow:
This sample detects faces in a video stream using DL Streamer's gvadetect element and saves frames containing detected objects to disk as JPEG images. It uses Gst.parse_launch to build and run the GStreamer pipeline from Python.

### DIFF
--- a/.github/actions/common/check-skip-ci/action.yml
+++ b/.github/actions/common/check-skip-ci/action.yml
@@ -28,8 +28,10 @@ runs:
     - name: Check if workflow_dispatch
       id: dispatch-check
       shell: bash
+      env:
+        EVENT_NAME: ${{ github.event_name }}
       run: |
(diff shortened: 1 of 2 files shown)


tjanczak:
Why are we doing this? If a sample is already in command line (bash) version, then it is more compact than it's Python counterpart. Why to make simple version more complex?

Please update existing command-line sample with
1) replace gvapython with regular GStreamer python plugin
2) replace OMZ models with corresponding models from HuggingFace

Do not convert command line sample to Python application; command line sample should still be invoked with gst-lanuch and one (or more) of elements in a command line can point to a python element.

tjanczak (changes requested):
do not change command line to python, instead upgrade command line samples to use regular python elements and not gvapython

PR #779, opened 2026-04-15
Added cross-stream grouping to batches and nireqs in optimizer

pbartosik:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


jmotow (approved):

scripts/optimizer/processors/inference.py
```diff
@@ -130,22 +130,44 @@ def __next__(self) -> list:
...
+        instance_ids = {}
+
         for idx, element in enumerate(self.pipeline):
             if "gvadetect" in element or "gvaclassify" in element:
+                (_, parameters) = parse_element_parameters(element)
+                instance_id = parameters.get("model-instance-id")
+                group_idx = 0
+
+                # if element has an instance id, get the batch group index
+                if instance_id:
```
oonyshch, line 153: using `if instance_id:` can treat empty strings as falsy, causing elements with `instance_id=""` to not be grouped together.

maybe use `if instance_id is not None:` instead to properly handle empty strings as valid grouping keys

scripts/optimizer/processors/inference.py
```diff
@@ -201,22 +222,44 @@ def __next__(self) -> list:
...
+        instance_ids = {}
+
         for idx, element in enumerate(self.pipeline):
             if "gvadetect" in element or "gvaclassify" in element:
+                (_, parameters) = parse_element_parameters(element)
+                instance_id = parameters.get("model-instance-id")
+                group_idx = 0
+
+                # if element has an instance id, get the nireq group index
+                if instance_id:
```
oonyshch, line 245: using `if instance_id:` will treat empty strings as falsy here imo

oonyshch (changes requested):
in the grouping logic for both `BatchGenerator` and `NireqGenerator`
the `if instance_id:` checks might fail for empty string values, causing elements with the same empty instance_id to be ungrouped when they should share the same batch/nireq group. 

upd: @walidbarakat corrected me right
LGTM then :)

walidbarakat (approved):
LGTM ... 

@pbartosik a reminder please to spend few sentences to explain the decision and impact of change. that would save time and effort for reviewers

@oonyshch instance grouping happens only when user explicitly set `model-instance-id` to be the same value/id, otherwise - if `model-instance-id` is not set or set to null - OV will set automatically the instance id to be unique. in other words, each instance id is unique and each instance/stream is a group consists of one member .. unless the user specify otherwise.

oonyshch (approved):

PR #793, opened 2026-04-21
fix instant EOS and tracking metadata copy failure in multi-branch pipelines

jmotow:
**Title: fix(gvametaaggregate): fix instant EOS and tracking metadata copy failure in multi-branch pipelines**

Description:
Fixes gvametaaggregate failing immediately with EOS when used in a multi-branch tee pipeline with gvatrack.

**Problem:**
Two bugs prevented gvametaaggregate from working in multi-branch pipelines:

Instant EOS — gst_aggregator_selected_samples() was not called before producing output, causing aggregator pads to enter an error state and the pipeline to terminate in ~0.05s instead of processing frames.

Tracking metadata relation copy failure — copy_all_gst_analytics_mtd() attempted to copy relations referencing tracking metadata that was intentionally skipped (not added to id_map). This caused gst_analytics_relation_meta_set_relation() to fail, returning GST_FLOW_ERROR.

**Changes:**
metaaggregate.c:

Added gst_aggregator_selected_samples() call in gst_gva_meta_aggregate_aggregate() before the EOS check, signaling the aggregator that samples have been selected for the current output segment.
Changed copy_all_gst_analytics_mtd() to log and skip (via continue) when a relation references a metadata ID not found in id_map, instead of returning FALSE.

**Testing:**
Tested with a two-branch tee pipeline (yolov4 + trafficcamnet) using gvametaaggregate, gvatrack, gvawatermark, and GPU inference with pre-process-backend=opencv. Pipeline completes successfully (exit code 0).

Note: The pipeline also requires vapostproc after queue in each tee branch to avoid a VA surface race condition (separate from this fix)


oonyshch (dismissed):

dmichalo (commented):
Code style fails.

oonyshch (approved):

dmichalo (approved):

PR #795, opened 2026-04-21
Improve performance on NPU related to segfault with dual inference pipeline on yolo11s

oonyshch:
## Description

When two `gvadetect` pipelines target `device=NPU` simultaneously, the NPU VCL compiler crashes with `ConvertVPUMI37XX2ELF failed: bad optional access` followed by SIGSEGV. This limitation is coming from the NPU compiler itself.

The current intent of this PR is to introduce clamping to batch-size of 1 for NPUs since it helps achieve better performance in DL Streamer.

### Reproduction

**Single pipeline:**

```bash
MODEL_PATH=/path/to/yolo11s/INT8/yolo11s.xml
VIDEO_PATH=/path/to/1192116-sd_640_360_30fps.mp4

gst-launch-1.0 \
  filesrc location=$VIDEO_PATH \
  ! qtdemux ! h264parse ! vaapidecodebin \
  ! vapostproc ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! gvadetect model=$MODEL_PATH pre-process-backend=va device=NPU batch-size=8 nireq=6 \
  ! gvafpscounter \
  ! fakesink sync=false
```

**Dual pipeline:**

```bash
MODEL_PATH=/path/to/yolo11s/INT8/yolo11s.xml
VIDEO_PATH=/path/to/1192116-sd_640_360_30fps.mp4

gst-launch-1.0 \
  filesrc location=$VIDEO_PATH \
  ! qtdemux ! h264parse ! vaapidecodebin \
  ! vapostproc ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! gvadetect model=$MODEL_PATH pre-process-backend=va device=NPU batch-size=8 nireq=6 \
  ! gvafpscounter \
  ! fakesink sync=false \
  filesrc location=$VIDEO_PATH \
  ! qtdemux ! h264parse ! vaapidecodebin \
  ! vapostproc ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! vapostproc \
  ! 'video/x-raw(memory:VAMemory)' \
  ! gvadetect model=$MODEL_PATH pre-process-backend=va device=NPU batch-size=8 nireq=6 \
  ! gvafpscounter \
  ! fakesink sync=false
```

**upd:** after thorough benchmarking on arl-h, and trying locking mutexes for npu the sigsegv described in does not reproduce on current main. dual gvadetect pipeline completes successfully for batch-sizes 1 through 8.  with issues from compiler observed for batch sizes of 4 and 8.
the proposed compile_mutex fix shows marginal performance benefit vs main. benchmarks are within noise; some cases are marginally slower with the lock.
batch-size 16/32 still fails, but it's an npu vcl compiler issue, escalating to NPU compiler team. 

benchmarks [fps] with previously tried fixes:
| batch | single fix | single main | dual fix | dual main |
|-------|-----------|-------------|----------|-----------|
| 1     | 90.49     | 90.18       | 88.95    | 88.98     |
| 2     | 29.58     | 29.67       | 32.25    | 38.09     |
| 4     | 31.40     | 31.23       | 19.74    | 20.05     |
| 8     | 26.16     | 25.94       | 30.69    | 30.71     |

so, at the end, either just clamping to 1, because this allows for the best performance in DL Streamer,
or figuring out better how to optimize in this case for throughput on NPUs
benchmarks  then everytime around ~90 fps

**upd2:** committed a workaround that maps batch size to nireq for async throughput

### How Has This Been Tested?

Locally on ARL-H with Intel Core Ultra 9 285H.

### Checklist:

### DIFF
--- a/src/gst/elements/gvafpscounter/fpscounter_c.cpp
+++ b/src/gst/elements/gvafpscounter/fpscounter_c.cpp
@@ -68,6 +68,7 @@ void fps_counter_create_readpipe(void *fpscounter, const char *pipe_name) {
         if (not fps_counters.count("readpipe")) {
             auto pipe_complete_lambda = [=]() {
                 fps_counter_eos(GST_ELEMENT_NAME(GST_BASE_TRANSFORM(fpscounter)));
+                // Pushing an EOS event downstream to signal that we're done.
                 bool handled = gst_pad_push_event(GST_BASE_TRANSFORM(fpscounter)->srcpad, gst_event_new_eos());
                 if (!handled)
                     throw std::runtime_error("FpsCounter ReadPipe: EOS event wasn't handled. Spinning...");
--- a/src/gst/elements/gvafpscounter/fpscounter_c.cpp
+++ b/src/gst/elements/gvafpscounter/fpscounter_c.cpp
@@ -41,10 +41,10 @@ void fps_counter_create_iterative(const char *intervals, bool print_std_dev, boo
 
 void fps_counter_create_average(unsigned int starting_frame, unsigned int interval) {
     try {
+        std::lock_guard<std::mutex> lock(channels_mutex);
         if (not fps_counters.count("average")) {
-            std::shared_ptr<FpsCounter> fps_counter = nullptr;
-            fps_counter = std::make_shared<IterativeFpsCounter>(starting_frame, interval, true, false, false);
-            fps_counters.insert({"average", fps_counter});
+            fps_counters.insert(
+                {"average", std::make_shared<IterativeFpsCounter>(starting_frame, interval, true, false, false)});
         }
     } catch (std::exception &e) {
         GVA_ERROR("Error during creation average fpscounter: %s", Utils::createNestedErrorMsg(e).c_str());
@@ -53,10 +53,9 @@ void fps_counter_create_average(unsigned int starting_frame, unsigned int interv
 
 void fps_counter_create_writepipe(const char *pipe_name) {
     try {
+        std::lock_guard<std::mutex> lock(channels_mutex);
         if (not fps_counters.count("writepipe")) {
-            std::shared_ptr<FpsCounter> fps_counter = nullptr;
-            fps_counter = std::make_shared<WritePipeFpsCounter>(pipe_name);
-            fps_counters.insert({"writepipe", fps_counter});
+            fps_counters.insert({"writepipe", std::make_shared<WritePipeFpsCounter>(pipe_name)});
         }
     } catch (std::exception &e) {
         GVA_ERROR("Error during creation writepipe fpscounter: %s", Utils::createNestedErrorMsg(e).c_str());
@@ -65,18 +64,17 @@ void fps_counter_create_writepipe(const char *pipe_name) {
 
 void fps_counter_create_readpipe(void *fpscounter, const char *pipe_name) {
     try {
+        std::lock_guard<std::mutex> lock(channels_mutex);
         if (not fps_counters.count("readpipe")) {
             auto pipe_complete_lambda = [=]() {
                 fps_counter_eos(GST_ELEMENT_NAME(GST_BASE_TRANSFORM(fpscounter)));
-                // Pushing an EOS event downstream to signal that we're done.
                 bool handled = gst_pad_push_event(GST_BASE_TRANSFORM(fpscounter)->srcpad, gst_event_new_eos());
                 if (!handled)
                     throw std::runtime_error("FpsCounter ReadPipe: EOS event wasn't handled. Spinning...");
             };
             auto new_message_lambda = [](const char *name) { fps_counter_new_frame(NULL, name, NULL); };
-            std::shared_ptr<FpsCounter> fps_counter = nullptr;
-            fps_counter = std::make_shared<ReadPipeFpsCounter>(pipe_name, new_message_lambda, pipe_complete_lambda);
-            fps_counters.insert({"readpipe", fps_counter});
+            fps_counters.insert({"readpipe", std::make_shared<ReadPipeFpsCounter>(pipe_name, new_message_lambda,
+                                                                                  pipe_complete_lambda)});
         }
     } catch (std::exception &e) {
         GVA_ERROR("Error during creation readpipe fpscounter: %s", Utils::createNestedErrorMsg(e).c_str());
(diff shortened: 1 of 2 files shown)


src/gst/elements/gvafpscounter/fpscounter_c.cpp
```diff
@@ -65,18 +64,17 @@ void fps_counter_create_writepipe(const char *pipe_name) {
 
 void fps_counter_create_readpipe(void *fpscounter, const char *pipe_name) {
     try {
+        std::lock_guard<std::mutex> lock(channels_mutex);
         if (not fps_counters.count("readpipe")) {
             auto pipe_complete_lambda = [=]() {
                 fps_counter_eos(GST_ELEMENT_NAME(GST_BASE_TRANSFORM(fpscounter)));
-                // Pushing an EOS event downstream to signal that we're done.
```
pbartosik, line 71: Is there a reason to delete this comment?

src/monolithic/inference_backend/image_inference/openvino/openvino_image_inference.cpp
```diff
@@ -1129,6 +1139,7 @@ class OpenVinoNewApiImpl {
                 if (_app_context) {
                     try {
                         dlstreamer::D3D11ContextPtr d3d11_ctx = dlstreamer::D3D11Context::create(_app_context);
+                        std::lock_guard<std::mutex> lock(compile_mutex());
```
walidbarakat: why we need to create a lock for the remote context for GPU?

src/monolithic/inference_backend/image_inference/openvino/openvino_image_inference.cpp
```diff
@@ -1144,6 +1155,7 @@ class OpenVinoNewApiImpl {
                 if (_app_context) {
                     try {
                         dlstreamer::VAAPIContextPtr vaapi_ctx = dlstreamer::VAAPIContext::create(_app_context);
+                        std::lock_guard<std::mutex> lock(compile_mutex());
```
walidbarakat: why we need to create a lock for the remote context for GPU?

src/monolithic/inference_backend/image_inference/openvino/openvino_image_inference.cpp
```diff
@@ -1071,14 +1078,17 @@ class OpenVinoNewApiImpl {
...
+        {
+            std::lock_guard<std::mutex> lock(compile_mutex());
+            if (_openvino_context) {
+                _compiled_model = core().compile_model(_model, _openvino_context->remote_context(), ov_params);
+            } else {
+                std::string formatted_device = _device;
+                if (_batch_timeout > -1) {
+                    formatted_device = fmt::format("BATCH:{}({})", _device, _auto_batch_num_requests);
+                }
+                _compiled_model = core().compile_model(_model, formatted_device, ov_params);
```
walidbarakat: here the lock is used for every device and mode, narrow it down to NPU only

src/monolithic/inference_backend/image_inference/openvino/openvino_image_inference.cpp
```diff
@@ -809,7 +816,7 @@ class OpenVinoNewApiImpl {
             try {
                 _batch_size = core().get_property(config.device(), ov::optimal_batch_size);
             } catch (...) {
-                _batch_size = 1; // Fallback if optimal batch size property is not supported
+                _batch_size = 1;
```
walidbarakat: why delete the comment here?

walidbarakat (dismissed):
over all i think we shouldn't handle the batch size as NPU plugin is take caring of it already 

[NPU plugin batching](https://docs.openvino.ai/2025/openvino-workflow/running-inference/inference-devices-and-modes/npu-device/batching-on-npu-plugin.html) 

Image

I understand that in case user don't use the same model instance id - which is something against DLS performance guide - the concurrent access to NPU plugin compiling can be the issue. but why we need the locks in FPS counter functionality ?

another question, how this bug was discovered? i tried to run these repro pipelines on my ARL machine and go no reproduction. 

note : please let's keep the commit atomic and focusing the needed change for the bug resolving.

src/gst/elements/gvafpscounter/fpscounter_c.cpp
oonyshch: no, reverted.

jmotow (dismissed):

tbujewsk (changes requested):
In general, any changes to _batch_size calculations should be limited to min.
This logic is specific to some limited scenarios and may simply does not apply to other.

tbujewsk:
> over all i think we shouldn't handle the batch size as NPU plugin is take caring of it already
> 
> [NPU plugin batching](https://docs.openvino.ai/2025/openvino-workflow/running-inference/inference-devices-and-modes/npu-device/batching-on-npu-plugin.html)
> 
> Image
> I understand that in case user don't use the same model instance id - which is something against DLS performance guide - the concurrent access to NPU plugin compiling can be the issue. but why we need the locks in FPS counter functionality ?
> 
> another question, how this bug was discovered? i tried to run these repro pipelines on my ARL machine and go no reproduction.
> 
> note : please let's keep the commit atomic and focusing the needed change for the bug resolving.

This is exactly the limitation that I mentioned about

walidbarakat:
> > over all i think we shouldn't handle the batch size as NPU plugin is take caring of it already
> > [NPU plugin batching](https://docs.openvino.ai/2025/openvino-workflow/running-inference/inference-devices-and-modes/npu-device/batching-on-npu-plugin.html)
> > Image
> > I understand that in case user don't use the same model instance id - which is something against DLS performance guide - the concurrent access to NPU plugin compiling can be the issue. but why we need the locks in FPS counter functionality ?
> > another question, how this bug was discovered? i tried to run these repro pipelines on my ARL machine and go no reproduction.
> > note : please let's keep the commit atomic and focusing the needed change for the bug resolving.
> 
> This is exactly the limitation that I mentioned about

Over all i don't like this PR ..

oonyshch:
> In general, any changes to _batch_size calculations should be limited to min. This logic is specific to some limited scenarios and may simply does not apply to other.

Okay, let's close then, and just wait for resolution on the NPU compiler's side, maybe? Then come up with a more fancy handling or this. I have submitted the information from my debugging to the NPU team, filed two issues.

PR #801, opened 2026-04-24
agent enhacement

jmotow:
Some coding agent enhancements.


tbujewsk (changes requested):
Please provide more justification for the changes. 
We should keep agent as simple as possible will all relevant info for handling properly the main DLS prompts.
Every agent related change must be preceded by tests (prompts tests).

PR #805, opened 2026-04-27
Run Unit Tests after building DLStreamer from sources on host.

nszczygl9:
### Description

Enabled Unit Tests execution on hosts after building DL Streamer from sources. Because there are some compilation problems on U22, it's only enabled on U24. 
For U22 CI uses external VM GH runner - DL Streamer is build without building Unit Tests.
For U24 CI uses self hosted U24 runner - DL Streamer and Unit Tests are build, Unit Tests are executed after building.

### Checklist:


tbujewsk (changes requested):
U22 path must be solved before merge

PR #811, opened 2026-04-30
Add Optimizer Unit Tests.

nszczygl9:
### Description

**_Added 7 more unit tests for optimizer which test functions like:_**
- iter_optimize_for_fps
- iter_optimize_for_streams
- get_optimal_pipeline
- get_baseline_pipeline
- optimize_for_streams
- optimize_for_fps
- set_allowed_devices
- set_sample_duration

Current execution of Unit Tests on main takes around **5 minutes.**
New execution with these new tests take around **16 minutes.**

### Checklist:


tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+        candidates = []
+        # Iterate through candidates and collect pipelines and their FPS
+        for candidate_result in optimizer.iter_optimize_for_fps(self.simple_pipeline):
+            if isinstance(candidate_result, tuple) and len(candidate_result) >= 2:
+                pipeline = candidate_result[0]
+                fps = candidate_result[1]
+                candidates.append((pipeline, fps))
+                print(f"Tested: {pipeline} @ {fps} FPS")
+            else:
+                continue
```
pbartosik:
```suggestion
        for pipeline, fps in optimizer.iter_optimize_for_fps(self.simple_pipeline):
            candidates.append((pipeline, fps))
            print(f"Tested: {pipeline} @ {fps} FPS")
```

pbartosik: is there a reason to check whether they are a tuple? You can just get the parts directly

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+        optimal_pipeline, optimal_fps, _ = optimizer.get_optimal_pipeline()
+        print(f"Optimal pipeline: {optimal_pipeline} @ {optimal_fps} FPS")
+        # Assert that the best candidate matches the optimal pipeline and FPS (allowing some tolerance for FPS)
+        self.assertEqual(best_candidate_pipeline, optimal_pipeline,
+                        f"Best candidate pipeline {best_candidate_pipeline} doesn't match "
+                        f"optimal pipeline {optimal_pipeline}")
+        # Allow a small tolerance for FPS comparison, since it can vary slightly due to system load and other factors
+        self.assertAlmostEqual(best_candidate_fps, optimal_fps, places=2,
+                            msg=f"FPS mismatch: candidate {best_candidate_fps} vs optimal {optimal_fps}")
+
```
pbartosik, line 59: Since you're testing `iter_optimize_for_fps` and `get_optimal_pipeline` in the same test, there shouldn't be any difference in fps here. For a difference in fps to arise, you would need to run a second optimization process. `get_optimal_pipeline` is expected to merely return the ebst bcadidate that the last optimization run has discovered.

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+    #     # Iterate through candidates and collect pipelines and their stream counts
+    #     for candidate_result in optimizer.iter_optimize_for_streams(self.simple_pipeline):
+    #         if isinstance(candidate_result, tuple) and len(candidate_result) >= 2:
+    #             pipeline = candidate_result[0]
+    #             stream_count = candidate_result[2]
+    #             fps_count = candidate_result[1]
+    #             candidates.append((pipeline, stream_count, fps_count))
+    #             print(f"Tested: {pipeline} @ {stream_count} streams @ {fps_count} FPS")
+    #         else:
+    #             continue
```
pbartosik:
```suggestion
    #     for pipeline, fps_count, stream_count in optimizer.iter_optimize_for_streams(self.simple_pipeline):
    #         candidates.append((pipeline, stream_count, fps_count))
    #         print(f"Tested: {pipeline} @ {stream_count} streams @ {fps_count} FPS")
```

pbartosik: same here, you can retrieve the parts of the tuple directly

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+
+    #     # We expect to have multiple candidates tested, at least more than 1
+    #     self.assertGreater(len(candidates), 1, 
+    #                     f"Expected more than 1 tested pipeline, got {len(candidates)}")
+        
+    #     # Find the candidate with the highest stream count
+    #     best_candidate = max(candidates, key=lambda x: x[1])
+    #     best_candidate_pipeline, best_candidate_streams, best_candidate_fps = best_candidate
+        
+    #     print(f"Best from candidates: {best_candidate_pipeline} @ {best_candidate_streams} streams @ {best_candidate_fps} FPS")
```
pbartosik: With `search_duration = 30` and `sample_duration = 10` you're likely to get only 3 results. By default Optimizer could be checking 3+ devices, 6 different batch sizes and 8 nireq numbers, so 17+ combinations for a pipeline with a single element. This means you're never going to reach higher stream numbers than 1. This will likely need to be a longer test.

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+        print(f"Optimal pipeline: {optimal_pipeline} @ {optimal_fps} FPS")
+        
+        # Check that the optimal pipeline matches the one returned by optimize_for_fps
+        self.assertEqual(optimal_pipeline, optimized_pipeline,
+                        f"Optimal pipeline {optimal_pipeline} doesn't match "
+                        f"optimized pipeline {optimized_pipeline}")
+        
+        # Allow a small tolerance for FPS comparison
+        self.assertAlmostEqual(optimal_fps, fps, places=2,
+                            msg=f"FPS mismatch: optimized {fps} vs optimal {optimal_fps}")
```
pbartosik, line 144: Again, there shouldn't be any difference between the values here

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
```diff
@@ -5,20 +5,271 @@
...
+        
+        # Assertions
+        # Long duration should take more time than short duration
+        self.assertGreater(elapsed_time_long, elapsed_time_short,
+                        f"Long duration ({elapsed_time_long:.1f}s) should be greater than short duration ({elapsed_time_short:.1f}s)")
+        
+        # More candidates should be tested with shorter duration (faster sampling)
+        self.assertGreaterEqual(len(candidates_short), len(candidates_long),
+                            f"Short duration should test more candidates: {len(candidates_short)} >= {len(candidates_long)}")
+
```
pbartosik: The amount of tested candidates will not differ since the `iter` functions do not care for passage of time. You will either have to let them run till finish (or set number of iterations) and compare time, or stop testing after a certain period and compare amount of tested candidates. You won't be able to check both.

pbartosik (changes requested):

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
nszczygl9: Exactly, there shouldn't be any difference in fps here. So this is an additional check if get_optimal_pipeline found the best pipeline with the best FPS. I would leave it like this...

tests/unit_tests/tests_gstgva/test_pipeline_optimizer.py
nszczygl9: but this test when run without any time restrictions takes more than 6h and causes timeout in CI...it can't be like that  https://github.com/open-edge-platform/dlstreamer/actions/runs/25117354905/job/73608189803

pbartosik: it doesn't have to run without any restrictions, but you should give it at least 15~20 minutes, otherwise the result will always be that max amount of streams which were achieved was 1

nszczygl9: so your suggestion is to just kill the execution after 15 mins and check if number of streams has changed in testes pipelines?

pbartosik: Yes, exactly.

pbartosik (dismissed):

pbartosik (dismissed):

pbartosik (dismissed):

pbartosik (approved):

tbujewsk (approved):

PR #823, opened 2026-05-06
Add support for GETI semantic segmentation

msmiatac:
### Description

This PR adds support for Geti semantic segmentation models in the tensor post-processing path. These models (like LiteHRNet) may emit per-class score maps that need post-processing before downstream components such as gvawatermark can consume them correctly. 

Previously, Geti models exported with model_type="Segmentation" were not mapped to a dedicated converter. This change introduces a new semantic_segmentation tensor converter, wires Geti/OpenVINO model metadata to it, and normalizes semantic-segmentation outputs to the existing semantic_mask tensor contract used by downstream consumers.

What changed

1. Added a new semantic_segmentation tensor converter.
2. Mapped Geti Segmentation models to that converter in the model API and post-processing alias resolution.
3. Implemented conversion of semantic-segmentation outputs into frame-level semantic_mask tensors.
4. Supported both:
        -   FP32 score maps shaped like [B,C,H,W], converted to class-index masks via per-pixel argmax
        -   integer mask outputs shaped like [B,H,W] or [B,1,H,W]


brmarkus:
Can you add a README (or update an existing one) and samples (sample command line, sample script), showing which models to use as examples (which `Geti models exported with model_type="Segmentation"`), where to get some, how to get some, please?

src/monolithic/gst/elements/gvawatermark/renderer/cpu/renderer_cpu.cpp
```diff
@@ -15,10 +15,10 @@ namespace {
 const int sigmaX = 15;
 const int sigmaY = 15;
 
-const std::vector<cv::Vec3b> PascalVoc21ClColorPalette = {
+const std::vector<cv::Vec3b> SemanticMaskColorPalette = {
     cv::Vec3b(0, 0, 0),       // background
-    cv::Vec3b(128, 0, 0),     // aeroplane
-    cv::Vec3b(0, 128, 0),     // bicycle
+    cv::Vec3b(0, 128, 0),     // first foreground class
+    cv::Vec3b(128, 0, 0),     // bicycle
```
mholowni: Why colors are swapped here?

src/monolithic/gst/inference_elements/common/post_processor/converters/to_tensor/semantic_segmentation.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <vector>
+
+using namespace post_processing;
+
+namespace {
+
+using Blob = InferenceBackend::Blob;
+using OutputBlob = InferenceBackend::OutputBlob;
+
+std::vector<size_t> getUnbatchedDims(const std::vector<size_t> &dims, size_t batch_size) {
```
mholowni, line 23: Maybe it would be better to return std::span to avoid copying?

src/monolithic/gst/elements/gvawatermark/renderer/cpu/renderer_cpu.cpp
msmiatac: The colors were changed because a green mask marks the objects from the main class better than the blue mask and the existing semantic_mask converter developed for DeeplabV3 that originally introduced PascalVoc21ClColorPalette already seems abandoned. But I will simply introduce another palette for the semantic_segmentation converter.

src/monolithic/gst/inference_elements/common/post_processor/converters/to_tensor/semantic_segmentation.cpp
msmiatac: The copied range is only the unbatched shape, typically 2 to 3 elements, so the cost is negligible. Returning a vector also avoids lifetime assumptions about GetDims storage that a span-based API would introduce.

mholowni (approved):

dmichalo (approved):

PR #838, opened 2026-05-13
fix: draw_face_attributes sample fails to build on Ubuntu 24 deb install due to missing OpenCV paths

OskarFiedot:
## Problem

When running the `draw_face_attributes` C++ sample on Ubuntu 24.04 with DL Streamer installed from Debian packages, the build fails with:

```
fatal error: opencv2/opencv.hpp: No such file or directory
```

**Reproduction command:**
```bash
/home/labrat/intel/dlstreamer/samples/gstreamer/cpp/draw_face_attributes/build_and_run.sh \
  "/tmp/video-examples/face-demographics-walking-2min.mp4  ! identity eos-after=1000 " \
  "json" "GPU" "/home/labrat/intel/dlstreamer/tests"
```

## Root cause

The `build_and_run.sh` script only passed OpenCV include/linker flags to CMake when `VERSION_ID` was `22.04`. On Ubuntu 24.04 (or any other version), these flags were omitted, so the compiler could not find the OpenCV headers installed in `/opt/opencv`.

## Changes

- Always pass OpenCV CMake flags (`-I/opt/opencv/include`, `-L/opt/opencv`) regardless of Ubuntu version
- Extract CMake OpenCV flags into a reusable bash array for clarity
- Add `set -euo pipefail` so the script fails fast on errors instead of silently continuing (e.g. running a non-existent binary after a failed build)
- Quote variables in the final `cp` command to correctly handle paths with spaces

### Checklist:


mholowni (dismissed):

samples/gstreamer/cpp/draw_face_attributes/build_and_run.sh
```diff
@@ -32,17 +34,15 @@ cd "${BUILD_DIR}" || exit
...
-    cmake3 "${BASE_DIR}"
+    cmake3 "${BASE_DIR}" "${CMAKE_OPENCV_FLAGS[@]}"
 fi
 
 make -j "$(nproc)"
 
 "${BUILD_DIR}"/draw_face_attributes -i "$FILE" -o "$OUTPUT" -d "$DEVICE"
-cp ${BUILD_DIR}/output.json $OUTPUT_DIRECTORY
\ No newline at end of file
+cp "${BUILD_DIR}/output.json" "$OUTPUT_DIRECTORY"
```
nszczygl9: add empty line

nszczygl9 (approved):

mholowni (approved):

PR #844, opened 2026-05-15
Add D3D11 GPU support for motion detect

yunowo:
### Description

Add D3D11 GPU support for motion detect

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


yangjianfeng1208 (approved):

qianlongding (approved):

src/monolithic/gst/elements/gvamotiondetect/gvamotiondetect_win.cpp
```diff
@@ -399,34 +590,65 @@ static void gst_gva_motion_detect_attach_metadata(GstGvaMotionDetect *self, GstB
...
-    } else {
-        return GST_FLOW_OK;
-    }
+
     int target_w = std::min(320, width);
     double scale = (double)target_w / (double)width;
     int small_w = target_w;
     int small_h = std::max(1, (int)std::lround(height * scale));
+
+    // Lazy one-shot peer query for the D3D11 device if set_context hasn't fired yet.
```
tbujewsk, line 623: Shouldn't this part (i.e. getting d3d11_device) be moved to gst_gva_motion_detect_start (or to gst_gva_motion_detect_class_init) ?

src/monolithic/gst/elements/gvamotiondetect/gvamotiondetect_win.cpp
```diff
@@ -307,16 +320,194 @@ static inline double md_iou(const MotionRectWin &a, const MotionRectWin &b) {
...
+            return FALSE;
+        try {
+            cv::directx::ocl::initializeContextFromD3D11Device(id3d);
+            self->cv_ocl_ready = TRUE;
+            GST_INFO_OBJECT(self, "cv::directx OpenCL context initialized from D3D11 device");
+        } catch (const cv::Exception &e) {
+            GST_WARNING_OBJECT(self, "initializeContextFromD3D11Device failed: %s", e.what());
+            return FALSE;
+        }
+    }
```
mholowni, line 462: Shouldn't it be moved to init/start? As i understand ocl context is created only once?

src/monolithic/gst/elements/gvamotiondetect/gvamotiondetect_win.cpp
yunowo: Linux version is doing the same thing, init device when first frame is coming.
Moving to gst_gva_motion_detect_class_init is not possible, it's too early, there is no upstream linked yet, the device query will never succeed. gst_gva_motion_detect_start is also too early, the caps are not set yet, we don't know if caps_is_d3d11 is true.

src/monolithic/gst/elements/gvamotiondetect/gvamotiondetect_win.cpp
yunowo: It's created once, but init/start is too early to get the d3d device.

mholowni (approved):

PR #858, opened 2026-05-22
Add gvaanalytics sample

mholowni:
### Description

Add sample for gvaanalytics.
This sample is showing example usage of gvaanalytics plugin - counting vehicles going from left to right and right to left.

N/A

### How Has This Been Tested?

### Checklist:


samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
```diff
@@ -0,0 +1,227 @@
...
+        print(f"Output will be saved to: {output_file}\n")
+    else:
+        output_sink = "videoconvert ! autovideosink"
+        print("Output will be displayed on screen (no file saving)\n")
+
+    # Build pipeline using parse_launch
+    pipeline_str = (
+        f"urisourcebin uri={video_file} ! "
+        f"decodebin3 ! "
+        f"gvadetect model={model_file} device=GPU batch-size=4 threshold=0.7 ! queue ! "
```
nszczygl9: Why device is hardcoded in pipeline?
This pipeline can be only run on GPU?

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Register the custom element so it can be used in parse_launch
+    if not Gst.Element.register(None, "vehicle_counter_text", Gst.Rank.NONE, VehicleCounterText):
+        sys.stderr.write("Failed to register vehicle_counter_text element\n")
+        sys.exit(1)
+
+    # Get path to tripwire config file
+    config_file = os.path.join(os.path.dirname(__file__), "tripwire-config.json")
+
+    # Build output sink based on whether file output is requested
```
nszczygl9, line 178: To be able to validate pipeline we need an option with json output file to compare real and golden metadata. Without json output pipeline cannot be validated in CI.

nszczygl9 (changes requested):

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
mholowni: no, configuration of device added

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
mholowni: added

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.sh
```diff
@@ -0,0 +1,26 @@
...
+# SPDX-License-Identifier: MIT
+# ==============================================================================
+
+set -e
+
+MODELS_PATH=${MODELS_PATH:-.}
+INPUT=${1:-https://videos.pexels.com/video-files/1192116/1192116-sd_640_360_30fps.mp4}
+MODEL=${2:-${MODELS_PATH}/public/yolo11n/FP16/yolo11n.xml}
+OUTPUT=${3:-/tmp/vehicle_counter_output.mp4}
+DEVICE=${4:-GPU}
```
nszczygl9: please add info about supported devices (CPU,GPU,NPU?) and add info what JSON_METADATA is;
maybe also a good idea is to validate MODELS_PATH somehow and displaying info when model cannot be found (?)

mholowni: added

samples/gstreamer/python/gvaanalytics_tripwire/tripwire-config.json
```diff
@@ -0,0 +1,17 @@
...
+                    "y": 360
+                },
+                {
+                    "x": 320,
+                    "y": 0
+                }
+            ]
+        }
+    ]
+}
```
nszczygl9: add empty line

nszczygl9 (dismissed):

samples/gstreamer/python/gvaanalytics_tripwire/tripwire-config.json
mholowni: added

nszczygl9 (dismissed):

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
```diff
@@ -0,0 +1,256 @@
...
+                        
+                        # Count the crossing
+                        if direction == 1:  # Left-to-right crossing
+                            crossing_count["left_to_right"] += 1
+                            crossing_count["total"] += 1
+                            print(f"Vehicle crossing (L→R) - {obj_type}! Total: {crossing_count['total']}")
+                        elif direction == -1:  # Right-to-left crossing
+                            crossing_count["right_to_left"] += 1
+                            crossing_count["total"] += 1
+                            print(f"Vehicle crossing (R→L) - {obj_type}! Total: {crossing_count['total']}")
```
walidbarakat: i have a doubt here. it seems that we're risking counting the detected - allowed - object(s) as long as they're in the frame even if it didn't cross the tripwire. for example if a `car` crosses the tripwire but the only other object on-frame is a `person`, the loop finds person first, which isn't in `_allowed_types`, and if `person` happens to be the only entry, `obj_type` stays "unknown". The crossing is still counted (the count logic runs unconditionally after the loop), but with a wrong label. If the intent was to filter, the crossing should have been skipped.

i think we can use `gvaanalytics` creates an explicit "RELATE_TO" that triggered it. from `gvaanalyticsimple.cpp` 

``` python
gst_analytics_relation_meta_set_relation(
    analytics_meta, GST_ANALYTICS_REL_TYPE_RELATE_TO, od_mtd.id, tripwire_mtd.id);
```

the code doesn't query this relation. 

i would suggest something like that

``` python
for mtd in relation_meta:
    if isinstance(mtd, DLStreamerMeta.TripwireMtd):
        success, tripwire_id, direction = mtd.get_info()
        if not success or direction == 0:
            continue

        # Walk all metadata to find the ODMtd related to THIS tripwire
        crossing_obj_type = None
        for related_mtd in relation_meta:
            if not isinstance(related_mtd, GstAnalytics.ODMtd):
                continue
            rel = relation_meta.get_relation(related_mtd.id, mtd.id)
            if rel & GstAnalytics.RelTypes.RELATE_TO:
                crossing_obj_type = GLib.quark_to_string(related_mtd.get_obj_type())
                break

        # Now filter on the ACTUAL crossing object's type
        if crossing_obj_type and crossing_obj_type.lower() not in self._allowed_types:
            continue  # Skip non-vehicle crossings

        # Count only verified vehicle crossings
        ...
```

samples/gstreamer/python/gvaanalytics_tripwire/README.md
```diff
@@ -0,0 +1,226 @@
...
+    {
+      "id": "vertical",
+      "points": [
+        {"x": 320, "y": 0},
+        {"x": 320, "y": 360}
+      ]
+    }
+  ]
+}
+```
```
walidbarakat, line 76: in `tripwire-config.json` defines points bottom to top `(320,360) -> (320,0)` i think the point order determines what "forward" (direction=1) means. This discrepancy means the README's direction labels ("left-to-right = direction 1") may be inverted relative to the actual config.

mholowni: Yes, you're right I've inverted the direction by mistake

mholowni: And, yes point order determines  what "forward" means

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
mholowni: https://github.com/open-edge-platform/dlstreamer/pull/858/commits/533ab2632118bfc4cffe54dcc6da28bf86556f06 - changed

oonyshch (dismissed):

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
```diff
@@ -0,0 +1,252 @@
...
+        
+        DLStreamerWatermarkMeta.text_meta_add(
+            buffer,
+            x=10, y=30,
+            text=counter_text,
+            font_scale=0.8,
+            font_type=0,  # cv::FONT_HERSHEY_SIMPLEX
+            r=0, g=255, b=255, thickness=1, draw_bg=True)
+
+        return Gst.FlowReturn.OK
```
walidbarakat, line 132: sorry if i wasn't clear, but this function seems to be important and need to have some thoughts : 

1) with using 
``` python
if crossing_obj_type and crossing_obj_type.lower() not in self._allowed_types: <- this line can shake the whole logic 
continue  # Skip non-vehicle crossings
```
if `crossing_obj_type` is None, which is an expected value in case meta relation look up fails. then this line will short circuit to false - `continue` will not fire - then the execution path will process to line 110 and so on. this is not valid behavior.

this should be :
``` python
if not crossing_obj_type or crossing_obj_type.lower() not in self._allowed_types:
      continue
```

walidbarakat (changes requested):

samples/gstreamer/python/gvaanalytics_tripwire/vehicle_counter.py
mholowni: https://github.com/open-edge-platform/dlstreamer/pull/858/commits/c8570cbd238eda7908221475aaebf9611dd52962

walidbarakat (approved):
looks ok

nszczygl9 (approved):

oonyshch (approved):

PR #861, opened 2026-05-22
Add YOLO classification support

msmiatac:
## Summary

Enable Ultralytics classification models (like `yolo26n-cls.pt`) in DL Streamer's automatic Model API conversion flow.

This adds support for `task: classify` that Ultralytics exports and routes it through DL Streamer's classification (`label`).

## What changed

- Added Ultralytics `classify` task support in the YOLO metadata-to-Model-API converter
- Mapped Ultralytics classification exports to DL Streamer's `label` model type instead of the YOLO detection-family suffix path
- Populated classification-specific Model API settings, including:
  - raw score output
  - classification preprocessing defaults
  - reshape from `imgsz` when available
 
## Why

Before this change, Ultralytics classification exports were not handled cleanly by the auto-conversion path even though the metadata already contains enough information to configure DL Streamer's classification pipeline.

This branch closes that gap by treating Ultralytics `classify` as a first-class task while still mapping it into DL Streamer's existing classification/label flow.


tbujewsk:
As discussed, pls review yaml parsing implementation and extend yolo_detection sample about classification case

tbujewsk (approved):

mholowni (approved):

src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
```diff
@@ -13,6 +13,45 @@
 
 namespace ModelApiConverters {
 
+namespace {
+
+std::vector<int> parseYoloImageSize(const nlohmann::json &yaml_json) {
+    std::vector<int> dims;
+    if (!yaml_json.contains("imgsz"))
+        return dims;
```
walidbarakat, line 21: it would be better to verify if field `imgsz` exist before calling the function. that would save the stack frame movement

src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
```diff
@@ -13,6 +13,45 @@
 
 namespace ModelApiConverters {
 
+namespace {
+
+std::vector<int> parseYoloImageSize(const nlohmann::json &yaml_json) {
```
walidbarakat, line 18: this function is consuming the output of `yaml2Json` which stores all values as strings. so `imgsz` either a string - represents an integer value - or array of strings. branch `is_number_integer` is pretty much a dead code. please rework the function accordingly 

plus, we need to make sure that string(s) represent only a valid integer dimensions

tbujewsk: thx, let's Maciej take a look on that next week

src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
tbujewsk: thx, let's Maciej take a look on that next week

PR #920, opened 2026-06-17
Add zero-shot OpenCLIP classification to gvaclassify

jelyoussefi:
### Description

Adds open-vocabulary (zero-shot) image classification to `gvaclassify`. Instead of a fixed classification head, `gvaclassify` runs a CLIP **image** encoder (vision tower + visual projection) and a new post-processing converter, `zeroshot_openclip`, scores the resulting image embedding by cosine similarity against precomputed **text-label embeddings** supplied at runtime as a `.safetensors` file. The label set lives outside the model, so classes are changed by regenerating the embeddings file (no retraining), and only the vision tower runs on the device, keeping the device graph static-shape (NPU-friendly).

- New converter `…/post_processor/converters/to_tensor/zeroshot_openclip.{h,cpp}`, selected when `gvaclassify zeroshot-embeddings-file=…` is set, or via `model-proc` (`"converter": "zeroshot_openclip"`).
- New `gvaclassify` properties: `zeroshot-embeddings-file`, `zeroshot-topk`.
- Embeddings loaded natively from `.safetensors` — no Python/PyTorch runtime dependency and no pickle code-execution risk on a file the design treats as untrusted.
- CLIP `logit_scale` (from the embeddings-file metadata, or a `model-proc` `logit_scale` param) is applied before the softmax so confidences are calibrated; class ranking is on raw cosine.
- Optional `model-proc` `unknown_threshold`: when the top-1 cosine is below it, the result is labelled `unknown` (`label_id = -1`) instead of being forced to the nearest class.
- Adds a sample, export/embedding tooling, a developer-guide page, and a unit test.

Co-developed with the original prototype authors (Wallace Lee, Kuan Heng Lee); ported to `main`.

None external. The converter parses the small `.safetensors` header with `nlohmann_json`, which is **already vendored in the repository** (`thirdparty/CMakeLists.txt`, FetchContent, MIT-licensed). This change only links the existing `nlohmann_json::nlohmann_json` target into `inference_elements`; no new third-party component is introduced. The `.safetensors` format itself is a data format, not a code dependency, read by a small in-tree header (`safetensors_reader.h`).

### How Has This Been Tested?

Unit test (no model download, no PyTorch):
- `test_postprocessing` includes `zeroshot_openclip_test.cpp`, which writes a small `.safetensors` fixture in C++ and checks class ranking, `logit_scale` calibration, and the `unknown_threshold` path. Run: `ctest -R test_postprocessing`.

End-to-end on Intel Core Ultra (Panther Lake), Ubuntu 24.04, OpenVINO/DL Streamer built from this branch, inference device CPU:
- CLIP `ViT-B/32` (`openai`) image encoder exported to OpenVINO IR via `samples/.../zero_shot_classification/tools/export_clip_vision_ov.py`; label embeddings built with `gen_label_embeddings.py`; classified with the sample pipeline (`inference-region=full-frame`).
- Labels: zebra, horse, dog, cat, car, bicycle, cup, bottle. Result: correct rank-1 on the test images (e.g. zebra → `zebra`, car → `car`). Reproduce with `samples/gstreamer/gst_launch/zero_shot_classification/zero_shot_classification.sh`.

### Checklist:


brmarkus:
What does it mean in terms of latency and throughput, could you compare legacy with the new converter, please? Any impact on resource usage memory and runtime)? Are all accelerator devices (still) supported?

jelyoussefi:
Thanks @brmarkus.
On latency and throughput, the converter itself adds very little. It does an L2 normalize, a cosine similarity against the class embeddings (num_classes x embedding_dim), a softmax and a top-k, all on the CPU and independent of the device. The main cost is the CLIP image encoder inference, which is the same with or without the converter. 
On memory and runtime, it actually uses fewer resources. The previous version loaded the embeddings with python3 / torch.load, so it needed Python and PyTorch in the runtime plus a subprocess. The new safetensors loader is plain C++ with no Python or torch dependency. The embeddings themselves are small (num_classes x embedding_dim x 4 bytes).
On devices, the converter runs on the CPU and does not depend on the device running the model, so nothing here is device specific.

jelyoussefi:
Thanks all. This is ready for review now. We'd like to use it for a customer enablement in Q3, so a review when you have capacity would be much appreciated. Could a maintainer also approve the CI workflows so the checks can run? Happy to address any feedback quickly.

tjanczak (commented):
Please note there is already a sample which does use CLIP-VIT with gvaclassify: https://github.com/open-edge-platform/dlstreamer/tree/main/samples/gstreamer/python/face_detection_and_classification

Is this PR adding a new model beyond CLIP-VIT?

jelyoussefi:
Thanks you for your prompt response, good to know about that sample.

That sample runs the fairface age classifier. It is a ViT fine-tuned with a fixed classification head, so it does closed-set classification: the classes (9 age groups) are trained into the model, and changing them means retraining or replacing the model. It is a supervised classifier, not a CLIP zero-shot setup.

This PR adds a different capability: **zero-shot, open-vocabulary classification**. gvaclassify runs the CLIP image encoder (no classification head), and the new zeroshot_openclip converter scores the image embedding by cosine similarity against text-label embeddings that are provided at runtime. In practice this means:

- The **classes are defined by text labels at runtime**, not trained into the model.
- You can add or change classes by swapping the embeddings file, **with no retraining**.

So it is not adding a new model beyond CLIP-ViT. It **adds a zero-shot classification path to gvaclassify** that does not exist today. The new parts are the zeroshot_openclip converter, the zeroshot-embeddings-file property, native safetensors loading of the embeddings (no Python or torch at runtime), and applying the CLIP logit_scale for calibrated scores.

tjanczak (commented):
Ok, makes sense. 
Can you move model download/preparation script to /scripts/download_models directory?
Can we merge this script with any existing model download scripts?

Please also note we want to move away from DLStreamer-specific model-proc file. Instead, we are using existing popular formats: TIMM metadata, HF config or Ultralytics. Can we download this model from HuggingFace and then use HuggingFace metadata format?
Alternatively, the model download script can use OpenVINO model API ("model_info" section inside model.xml).

jelyoussefi:
Thanks for the guidance. I reworked the PR along these lines and pushed the update.

- Moved the model preparation into scripts/download_models and reused the existing Hugging Face CLIP path. download_hf_models.py now exports the projected CLIP image embedding for zero-shot when called with --extra_args --zeroshot; the existing clip_token export is unchanged. I added clip_text_embeddings.py there for the text-label embeddings, and switched both helpers from open_clip to transformers to match the other scripts.
- Dropped the DL Streamer model-proc file. Image preprocessing (CLIP mean/std, RGB, center crop) now comes from the model_info section of model.xml, the same mechanism the existing CLIP export uses. The converter is selected by the zeroshot-embeddings-file property.
- Moved logit_scale and the optional unknown_threshold into the embeddings-file metadata, so the converter no longer reads any model-proc output params.

Preparing the two artifacts with the same CLIP model:

  python3 download_hf_models.py --model openai/clip-vit-base-patch32 --extra_args --zeroshot
  python3 clip_text_embeddings.py --model openai/clip-vit-base-patch32 --labels labels.txt --output labels.safetensors

I built and installed this on an Intel platform. gst-inspect-1.0 gvaclassify lists the zeroshot-embeddings-file and zeroshot-topk properties, and the postprocessing unit test passes, including the zero-shot cases that cover class ranking, logit_scale calibration from the file metadata, and the unknown_threshold rejection path now read from that metadata.

docs/user-guide/dev_guide/gvaclassify_zeroshot_openclip_design.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+Only the CLIP vision tower runs on the inference device. The similarity, temperature scaling and
+top-k are cheap host-side operations, which keeps the device graph static-shape (important for NPU).
+
+## Selecting zero-shot mode
+
+Either is sufficient:
+
+- Set `gvaclassify zeroshot-embeddings-file=<labels>.safetensors`. With no `model-proc` converter
```
tjanczak, line 29: I'd just keep first option only, as discussed earlier we want to deprecate proprietary model-proc.... and the first option is very easy to use, so it would be preferred over additional model-proc. 
What about model file? We still need to set 'gvaclassify model=<path to model file>', right?

samples/gstreamer/gst_launch/zero_shot_classification/zero_shot_classification.sh
```diff
@@ -0,0 +1,62 @@
...
+
+for f in "$MODEL" "$EMBEDDINGS"; do
+  if [ ! -f "$f" ]; then
+    echo "Missing $f. Prepare the artifacts first (see the header of this script / README.md)."
+    exit 1
+  fi
+done
+
+if [[ "$INPUT" == *.jpg || "$INPUT" == *.jpeg || "$INPUT" == *.png || "$INPUT" == *.bmp ]]; then
+  SOURCE="filesrc location=${INPUT} ! decodebin3 ! videoconvert"
```
tjanczak, line 49: please remove 'videoconvert' from the pipeline, it should not be needed... and 'videoconvert' does GPU->CPU copy (this element does not support VAMemory)

scripts/download_models/requirements.txt
```diff
@@ -1,2 +1,3 @@
 ultralytics==8.4.3
 optimum-onnx==0.1.0
+safetensors
```
tjanczak, line 3: can you please pin specific version you were testing with?

tjanczak (dismissed):
Looks good, just added few minor comments.

msmiatac:
Two design points I'd like us to align on before merging. Both are about keeping the converter layer clean and making CLIP model selection unambiguous.

**1. Converters should be I/O-free: data in → tensors out.**

The codebase already establishes the convention that files become data in the post-processor layer, and converters stay pure (e.g. loadLabelsFromFile). The new converter breaks that: its constructor calls load_safetensors_matrix(), reads logit_scale / unknown_threshold from the safetensors metadata, and L2-normalizes the class rows - so it carries file-format knowledge, not just computation.

**Recommendation:** Move all safetensors handling out of the converter and into a loadEmbeddingsFromFile() in the post-processor setup, as a sibling to loadLabelsFromFile(). It should parse the matrix, L2-normalize the class rows, and read logit_scale / unknown_threshold from the file metadata, producing a ready-to-use struct (e.g. ZeroShotEmbeddings) passed through Initializer. That leaves the converter doing only cosine / softmax / top-k on ready data - no file, no format, no metadata keys. (Normalizing the incoming image embedding at runtime stays in the converter, since that's per-frame data, not file I/O.)

**2. Two CLIP converters share one model_type, which is confusing.**

We already have clip_token_converter, which emits the unprojected embedding - the vision tower's native output (pooler_output, or the raw CLS token from last_hidden_state) - intended for **image-to-image** comparisons performed _outside_ the DL Streamer pipeline. The new converter emits projected image embeddings and compares them against projected text embeddings for zero-shot classification.

So we now have two CLIP converters performing different tasks, both identified by the same model_type (clip_token) and differentiated at runtime only by a gvaclassify property. This is easy to misconfigure.

**Recommendation:** Rename zeroshot_openclip → clip_zeroshot, and emit a distinct model_type (e.g. clip_zeroshot) at model-conversion time. This makes converter selection deterministic from the model itself rather than from a runtime property, and lets the pipeline reject a mismatched combination - e.g. an unprojected clip_token model paired with a text-embeddings bank - instead of silently computing cosines across incompatible vector spaces.

docs/user-guide/dev_guide/gvaclassify_zeroshot_openclip_design.md
wallacezq: yes, gvaclassify model= must be set. When you mentioned model-proc is to be deprecated, does it affect the pre/post processor logic ? Ideally we still want to perform some pre/post processing function like converting the raw embedding tensor from the model output to human-readable zero-shot classification result.

samples/gstreamer/gst_launch/zero_shot_classification/zero_shot_classification.sh
wallacezq: ok, will do that

scripts/download_models/requirements.txt
wallacezq: ok will do that

msmiatac:
@copilot resolve the merge conflicts in this pull request

dmichalo:
Hello, please fix code-style and documentation issues. Additionally, one functional test failed, but that seems not as real issue.

docs/user-guide/dev_guide/gvaclassify_zeroshot_openclip_design.md
jelyoussefi: To answer your question: deprecating model-proc does not remove any pre/post-processing. The clip_zeroshot converter still runs and does exactly that - it turns the raw image-embedding tensor from the model output into the human-readable zero-shot result (L2 normalize, cosine similarity against the text-label embeddings, logit_scale, softmax, top-k). What changed is only where the configuration comes from: image preprocessing (CLIP mean/std, resize, color) now rides in the model_info section of model.xml instead of a model-proc JSON, and the converter is selected from the model's model_type. The processing logic is intact; we only dropped the proprietary JSON file.

jelyoussefi: To answer the pre/post-processing question: deprecating model-proc does not remove any pre/post-processing. The clip_zeroshot converter still runs and does exactly that - it turns the raw image-embedding tensor from the model output into the human-readable zero-shot result (L2 normalize, cosine similarity against the text-label embeddings, logit_scale, softmax, top-k). What changed is only where the configuration comes from: image preprocessing (CLIP mean/std, resize, color) now rides in the model_info section of model.xml, and the converter is selected from the model_type (clip_zeroshot). gvaclassify model= is still required and points at the CLIP image-encoder IR. The processing logic is intact; we only dropped the proprietary JSON file.

scripts/download_models/requirements.txt
jelyoussefi: Done - pinned to safetensors==0.7.0 (the version tested). Note a recent main merge split this file per tool, so it now lives in scripts/download_models/requirements_download_hf_models.txt, which the CLIP/HF export path uses.

jelyoussefi:
Code-style and documentation issues are addressed in the latest commits. The functional test failure looks unrelated to this change - could you confirm whether it's a known flaky/unrelated case?

jelyoussefi:
Could a maintainer approve the CI workflows so the required checks can run? The branch is up to date with main and the review comments are addressed. Happy to fix anything the checks surface.

dmichalo (approved):

tbujewsk (approved):

PR #927, opened 2026-06-19
Add GstAnalytics segmentation metadata support

OskarFiedot:
### Description

This change integrates `GstAnalyticsSegmentationMtd` (semantic and instance
segmentation) into DL Streamer, mirroring the existing analytics-metadata
bridging already in place for keypoints, object detection and classification.

It introduces a dedicated `"segmentation"` tensor type and routes segmentation
results through the standard GStreamer Analytics metadata model
(`GstAnalyticsRelationMeta`, GStreamer 1.26+), so semantic and instance masks
are produced, consumed and serialized consistently across the pipeline. As part
of this work the previously separate `semantic_mask` path is consolidated into a
single `semantic_segmentation` path (one converter, one tensor format, one color
palette).

Main changes:

- **New tensor type** — add `GST_ANALYTICS_SEGMENTATION_2_TENSOR` (`"segmentation"`)
  in [tensor.h](include/dlstreamer/gst/videoanalytics/tensor.h) and implement
  bidirectional conversion:
  - `convert_to_meta` (legacy → analytics): semantic `int64` class maps are
    stored as a `GstAnalyticsSegmentationMtd` (`GRAY8` / `GRAY16_LE` depending on
    the max class id) tagged with the plain model name; instance per-ROI masks
    are stored as a `GstAnalyticsTensorMtd` (`FP32` soft mask) tagged
    `<model_name>/instance_segmentation`.
  - `convert_to_tensor` (analytics → legacy): a `GstAnalyticsSegmentationMtd`
    always reconstructs a semantic-segmentation tensor (a `SegmentationMtd`
    always represents semantic segmentation), while a
    `<model>/instance_segmentation` `GstAnalyticsTensorMtd` reconstructs the
    instance mask. The model name is preserved as the semantic tag in both
    directions.
- **Converters** — tag segmentation tensors with the new type in
  [mask_rcnn.cpp](src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/mask_rcnn.cpp),
  [yolo_v8.cpp](src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/yolo_v8.cpp),
  [yolo_v26.cpp](src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/yolo_v26.cpp) and
  [semantic_segmentation.cpp](src/monolithic/gst/inference_elements/common/post_processor/converters/to_tensor/semantic_segmentation.cpp).
- **Semantic-segmentation consolidation** — remove the legacy `semantic_mask`
  path and merge it into the unified `semantic_segmentation` one:
  - delete the `SemanticMaskConverter`
    ([semantic_mask.{h,cpp}](src/monolithic/gst/inference_elements/common/post_processor/converters/to_tensor/blob_to_tensor_converter.cpp))
    and its registration in `BlobToTensorConverter::create`;
  - drop the `TENSOR_FORMAT_SEMANTIC_MASK` format constant and every branch that
    special-cased it in [tensor.h](include/dlstreamer/gst/videoanalytics/tensor.h)
    and [gstgvawatermarkimpl.cpp](src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp);
  - remove the `SemanticMaskPalette` enum and the per-mask `palette` field from
    `render::SemanticSegmantationMask`
    ([render_prim.h](src/monolithic/gst/elements/gvawatermark/renderer/render_prim.h));
    the renderer now always uses the single `PascalVoc21` palette;
  - map the old `semantic_mask` converter name to `semantic_segmentation` in
    `checkOnNameDeprecation`
    ([blob_to_meta_converter.cpp](src/monolithic/gst/inference_elements/common/post_processor/blob_to_meta_converter.cpp))
    for backward compatibility, and tag the deeplabv3 model accordingly in
    [download_public_models.sh](samples/download_public_models.sh).
- **Deduplication** — skip legacy segmentation tensors when iterating tensors and
  emit the analytics-sourced ones instead in
  [region_of_interest.h](include/dlstreamer/gst/videoanalytics/region_of_interest.h),
  [video_frame.h](include/dlstreamer/gst/videoanalytics/video_frame.h) and the
  gvametaconvert JSON converter
  ([jsonconverter.cpp](src/monolithic/gst/elements/gvametaconvert/jsonconverter.cpp)).
  `VideoFrame::get_tensors` additionally skips the internal transcription
  descriptor cls mtd so it is not surfaced as a frame-level tensor result.
- **Frame-level tensor handling** — rework `VideoFrame::add_tensor` to accept an
  externally-owned `Tensor` and write both the legacy `GstGVATensorMeta` (as a
  copy) and the matching frame-level analytics mtd (not attached to any
  `ODMtd`), mirroring `RegionOfInterest::add_tensor`. The ownership contract is
  now documented on both overloads (`VideoFrame` copies the structure and the
  caller keeps ownership; `RegionOfInterest` takes ownership). `gvaattachroi` is
  updated to route frame-level tensors through the new API — dropping the manual
  `convert_to_meta` that produced a duplicated analytics entry and the redundant
  per-object `objects` branch — and no longer needs the `GstBuffer` argument
  ([attachroi.cpp](src/monolithic/gst/elements/gvaattachroi/attachroi.cpp)).
- **Metadata aggregation** — `gvametaaggregate` now copies segmentation and
  tensor analytics mtd across aggregated frames
  ([metaaggregate.c](src/monolithic/gst/elements/gvametaaggregate/metaaggregate.c)).
- **Rendering fixes** — in
  [renderer_cpu.cpp](src/monolithic/gst/elements/gvawatermark/renderer/cpu/renderer_cpu.cpp):
  - `gvawatermark` NV12 instance-mask rendering now also blends the Y plane,
    producing correct mask colors;
  - semantic-mask colors are now correct for every negotiated pixel format: the
    RGB-ordered palette is routed through the renderer's per-format
    `_color_converter` (BGR swap for BGR-family buffers, identity for RGB),
    instead of being written raw and swapping red/blue on BGR buffers.
- **GIR script hardening** — [fix_gir_mtd_fields.py](scripts/fix_gir_mtd_fields.py)
  now parses GIR files via `defusedxml` and limits MTD records to
  `ZoneMtd` / `TripwireMtd`.
- **Documentation** — update the metadata developer guide to cover segmentation:
  extend the element I/O table with `GstAnalyticsSegmentationMtd` /
  `GstAnalyticsTensorMtd` ([metadata.md](docs/user-guide/dev_guide/metadata.md))
  and add semantic-segmentation (frame-level and per-ROI) and instance-
  segmentation sections, including the semantic-tag conventions
  ([metadata_analytics.md](docs/user-guide/dev_guide/metadata_analytics.md)).

Fixes # (issue)

No new 3rd party dependencies are introduced.

- `gstreamer-video-1.0` (GStreamer, LGPL-2.1+) is added as a link dependency for
  the `gstvideoanalyticsmeta` unit-test target. It is already part of the
  GStreamer stack used by DL Streamer and is only used here to access
  `GstVideoMeta` / video-format helpers when validating segmentation masks.
- `defusedxml` is already a project dependency; the GIR fix-up script is updated
  to use it for safe XML parsing instead of adding anything new.

### How Has This Been Tested?

New unit tests were added:

- [tensor_convert_test.cpp](tests/unit_tests/check/components/gstvideoanalyticsmeta/tensor_convert_test.cpp):
  10 segmentation conversion tests (7 in `SegmentationConvertToMetaTest`,
  3 in `SegmentationConvertToTensorTest`) plus 22 `SemanticTagTest` cases,
  covering:
  - semantic conversion success and correct `GstVideoMeta` format/size,
  - region count equal to the number of unique classes,
  - instance conversion (bbox location + region count) and rejection of
    non-integer (`FP32`) instance masks,
  - semantic / instance / `GRAY16_LE` round-trips and semantic-tag
    preservation.
- [metaaggregate_copy_test.cpp](tests/unit_tests/check/components/metaaggregate_copy/metaaggregate_copy_test.cpp):
  49 `MetaaggregateCopyTest` cases validating that segmentation and tensor
  analytics mtd are copied correctly during meta aggregation.

Validation:

- the new segmentation, semantic-tag and metaaggregate-copy tests pass,
- existing tests pass with no regressions,
- the `elements` (gvawatermark + gvametaconvert) and `inference_elements`
  targets compile cleanly.

The semantic-mask color fix was additionally validated end-to-end: rendering the
same deeplabv3 pipeline through the `gvawatermark` bin with `BGRA` now matches a
known-correct `RGBA` render (mean absolute pixel diff 1.35 vs 8.62 before the
fix).

### Checklist:


include/dlstreamer/gst/videoanalytics/tensor.h
```diff
@@ -919,6 +1066,222 @@ class Tensor {
...
+                tensor.set_data(mask.data(), mask.size() * sizeof(float));
+
+                return tensor.gst_structure();
+            }
+
+            // Generic raw tensor reconstruction: copy the payload verbatim and map the GstTensor
+            // data type to a GVA precision. Only fields that come directly from the analytics meta
+            // (precision, dims, payload and semantic tag) are restored.
+            Precision precision = Precision::UNSPECIFIED;
+            switch (gtensor->data_type) {
```
tjanczak: Why do we need such a big case statement; don't we have a generic function which does mapping?
If not, please add it as generic function and implement as a lookup table, rather than series of switch/case statements.

include/dlstreamer/gst/videoanalytics/video_frame.h
```diff
@@ -449,15 +474,33 @@ class VideoFrame {
...
                 if (mt == gst_analytics_cls_mtd_get_mtd_type()) {
+                    // Skip class descriptor metadata (used internally for label_id lookup)
                     gchar *tag = gst_analytics_mtd_get_semantic_tag(&mtd);
                     if (tag) {
                         bool is_descriptor = (strcmp(tag, "class_descriptor") == 0);
                         g_free(tag);
                         if (is_descriptor)
                             continue;
                     }
+                    // Skip the transcription descriptor cls mtd (label="transcription") added by
```
tjanczak, line 487: Why skipping audio transcription? It should be handled as any other frame-level classification. 
Let's avoid workarounds/special cases like that.

tjanczak (dismissed):

include/dlstreamer/gst/videoanalytics/video_frame.h
OskarFiedot: The only metadata we are omitting here is the descriptor type (cls mtd with a single ‘transcription’ label and no analytical data), just as a few lines above we omitted the descriptor we use for the label id

tbujewsk (approved):

mholowni (approved):

dmichalo (approved):

PR #957, opened 2026-07-03
[GST WATERMARK] support gvawatermark blurring on GPU

walidbarakat:
### Description

Enable GaussianBlur on VA surface (GPU) path in gvawatermark, allowing object blurring without falling back to CPU when the pipeline uses hardware-accelerated decode/encode.

### How Has This Been Tested?

using gstreamer pipeline :

gst-launch-1.0 filesrc location=`input_video.mp4` ! parsebin ! vah264dec ! gvadetect model=`model.xml` device=GPU pre-process-backend=va-surface-sharing ! gvawatermark device=GPU displ-cfg="enable-blur=true" ! vah264enc ! h264parse ! mp4mux ! filesink location=`output_video.mp4`

### Checklist:

### DIFF
--- a/src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
+++ b/src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
@@ -134,7 +134,7 @@ struct Impl {
     bool extract_primitives(GstBuffer *buffer);
     int get_num_primitives() const;
     bool render(GstBuffer *buffer);
-    bool render_va(cv::Mat *buffer);
+    [[nodiscard]] bool render_va(cv::Mat *overlay, cv::UMat *frame);
     const std::string &getBackendType() const {
         return _backend_type;
     }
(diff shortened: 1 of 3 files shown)


src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
```diff
@@ -630,25 +630,30 @@ static GstFlowReturn gst_gva_watermark_impl_transform_ip(GstBaseTransform *trans
...
-                    GST_INFO_OBJECT(gvawatermark, "Allocated CPU overlay (%dx%d CV_8UC4)", width, height);
+                    GST_INFO_OBJECT(gvawatermark, "Allocated CPU overlay (%dx%d CV_8UC3)", width, height);
                 }
 
                 // Clear only once per frame (fast memset on host)
                 gvawatermark->overlay_cpu.setTo(cv::Scalar(0, 0, 0, 0));
 
-                cv::UMat u;
-                cv::va_intel::convertFromVASurface(gvawatermark->va_dpy, sid, cv::Size(width, height), u);
+                if (!cv::ocl::useOpenCL()) {
```
brmarkus, line 639: `useOpenCL`: It checks initialization, not hardware. It returns true if OpenCL has been successfully initialized and is globally enabled within the binary. It means the framework is active. It confirms that OpenCV can route processing through OpenCL Transparent API instead of the standard CPU fallback.
But it's not a guarantee that OpenCL processing happens on a GPU.

brmarkus: cv::ocl::Device device = cv::ocl::Device::getDefault();

if (device.type() == cv::ocl::Device::TYPE_GPU) {
    std::cout << "Success: OpenCL is using the GPU!\n";
    std::cout << "Device Name: " << device.name() << "\n";
} else if (device.type() == cv::ocl::Device::TYPE_CPU) {
    std::cout << "Notice: OpenCL is initialized, but running on the CPU.\n";
} else {
    std::cout << "OpenCL is using an alternative device type.\n";
}

walidbarakat: you have a valid point here, but at line 472 `cv::va_intel::ocl::initializeContextFromVA(self->va_dpy, /*interop*/ true);` this function creates the OpenCL context using the `cl_intel_va_api_media_sharing` extension, which requires the GPU OpenCL device that shares memory with the VA display. It cannot produce a CPU OpenCL device — there's no VA-API/CPU-OpenCL interop path. After this call, `cv::ocl::Device::getDefault().type()` will always be TYPE_GPU

brmarkus: Thank you for the clarification

src/monolithic/gst/elements/gvawatermark/gstgvawatermarkimpl.cpp
```diff
@@ -901,15 +906,43 @@ bool Impl::render(GstBuffer *buffer) {
...
+    bool has_overlay_prims = false;
+    for (auto it = prims.begin(); it != prims.end();) {
+        if (std::holds_alternative<render::Blur>(*it)) {
+            const auto &blur = std::get<render::Blur>(*it);
+            cv::Rect r = blur.rect;
+            if (r.width <= 0 || r.height <= 0) {
+                it = prims.erase(it);
+                continue;
+            }
+            // Adaptive kernel: ~1/3 of ROI dimension, minimum 7, must be odd
```
tbujewsk: Let's do not keep the same logic in two places, use here `renderer_cpu.cpp::computeBlurKernelSize()`

So:
- move the `computeBlurKernelSize` out of the anonymous namespace {} block in `renderer/cpu/renderer_cpu.cpp`
- add `#include <opencv2/core.hpp>` and a forward declaration: of `cv::Size computeBlurKernelSize` in `renderer_cpu.h`
- include the `renderer/cpu/renderer_cpu.h` in `gstgvawatermarkimpl.cpp`

walidbarakat: i got your point. i just have committed a classic DRY sin :D 

however, i'm concerned about your suggestion that it may lead to more subtle issue with mixing the renderer-layer `renderer_cpu.h` with element-layer `gstgvawatermarkimpl.cpp` let's not do that. as it it can cause a coupling direction. however, you've the right point. so i suggest to move `computeBlurKernelSize` to `render_prim.h` to logically group it with Blur structure definition and as an already shared header between both CPU and GPU execution path.

tbujewsk (approved):

mholowni (approved):

PR #988, opened 2026-07-17
Optimizer: added power based optimization

pbartosik:
### Description

1. Added the option to optimize based on power draw to the optimizer library
2. Added the option to optimize based on power draw to the optimizer CLI tool
3. Power information is pulled from [Metrics Manager](https://github.com/open-edge-platform/edge-ai-libraries/tree/main/microservices/metrics-manager)
4. fps or power limits can now be configured for all optimization types

1. [prometheus-client](https://github.com/prometheus/client_python) (Apache-2.0)
2. [requests](https://github.com/psf/requests) (Apache-2.0)

### How Has This Been Tested?

### Checklist:


dmichalo:
While we are installing requirements.txt inside a dev container, its size growths. When this step will be approved and PR will be ready to merge - we will increase expected size of dev containers and rerun what's required.

walidbarakat (dismissed):

mholowni (dismissed):

docker/ubuntu/ubuntu24.Dockerfile
```diff
@@ -444,6 +444,9 @@ RUN \
        usermod -a -G video dlstreamer && \
        chown -R dlstreamer:dlstreamer /home/dlstreamer
 
+# Install python dependencies
+RUN pip3 install --no-cache-dir --break-system-packages --ignore-installed -r "${DLSTREAMER_DIR}/requirements.txt"
```
tbujewsk, line 448: Is it needed here since DLS requirements are also installed later Ln586:
`pip3 install --no-cache-dir --break-system-packages --ignore-installed -r /opt/intel/dlstreamer/requirements.txt &&` ?

scripts/optimizer/__main__.py
```diff
@@ -144,6 +158,27 @@ def main() -> int:
                 json_result["optimal"] = {"pipeline": best_pipeline, "metrics": best_result}
                 _display_summary_fps(best_pipeline, best_result["fps"], base_pipeline, base_result["fps"])
 
+            case "power":
+                json_result["mode"] = "power"
+                json_result["candidates"] = []
```
tbujewsk, line 163: Add check if both params (power_limit && power_metrics_endpoint) are set before running iter_optimize_for_power()

docker/ubuntu/ubuntu24.Dockerfile
pbartosik: Ln586 installs requirements in the final image, which is necessary to ship a working optimizer. Ln448 modified here installs requirements inside of the dev image, which is the image used for running unit tests. With the way our dockerfiles are setup, those two images are basically independent of each other.

scripts/optimizer/__main__.py
pbartosik: A power limit wouldn't be meaningful here since this mode tries to achieve lowest power draw anyway. I have added a check for the metrics endpoint inside of the `iter_optimize_for_power` and `optimize_for_power` methods so that it's performed at the library level.

tbujewsk (approved):

mholowni (approved):

PR #989, opened 2026-07-17
Add gvaupscale element for neural image upscaling

msmiatac:
### Description

Adds a new GStreamer element, gvaupscale, that performs neural super-resolution on video frames using an OpenVINO™ model. Unlike the existing gva* analytics elements — which attach metadata to a passthrough frame — gvaupscale emits new, larger frames (e.g. 2×). It is therefore a size-changing GstVideoFilter whose src caps differ from its sink caps.

This is the first DL Streamer element whose output is a transformed image rather than inference metadata, extending the framework beyond pure video analytics into generative/enhancement use cases.

### Motivation
Super-resolution is a common video-enhancement need (upscaling low-res streams for display, archival, or downstream analytics), but neither DL Streamer nor comparable frameworks ship a ready-made element for it — it has always required a custom plugin. This adds a first-class, OpenVINO-backed upscaler.

### Design
To avoid duplicating DL Streamer's optimized, parametrized pipeline, gvaupscale reuses the existing InferenceBackend::ImageInference engine (the same one gvainference drives) for preprocessing + inference. The only element-specific logic is writing the network's output tensor into a new, larger output frame.

- Base class: GstVideoFilter (GstBaseTransform). transform_caps scales width/height by the scale factor (sink→src) so the pipeline negotiates genuinely larger frames downstream.
- Preprocessing from model_info: color order and input normalization are read from the model's model_info rt_info (reverse_input_channels, scale_values/mean_values) via GetModelInfoPreproc, following the OpenVINO Model API convention — no hardcoded normalization.
- Dynamic models: fully-convolutional SR models (dynamic H/W) are statically reshaped to the input frame size at negotiation time.
- Output: the output tensor (float) is denormalized clip(x, 0, 1) · 255 → 8-bit and written into the output frame. This experimental version assumes Real-ESRGAN-style [0,1] output; additional model_info fields can be added later for models that diverge.

### Properties

- [ ] model	(string)	Path to the OpenVINO super-resolution model (.xml). Required.
- [ ] device	(string)	Inference device (CPU, GPU, NPU, AUTO, …).
- [ ] scale	(double)	Output/input resolution ratio 

Pads are video/x-raw, format=BGR (system memory), consistent with the other inference elements.

### Limitations / follow-ups (documented in the README)

- One synchronous inference per frame (no async nireq pipelining yet).
- System-memory preprocessing only (no VA-API surface sharing / zero-copy yet).
- Tiled inference for very large frames not yet implemented.
- Output denormalization currently assumes Real-ESRGAN-style [0,1] output; a dedicated model_info/property may be added for models that differ.


brmarkus:
Which models (sort of "Real-ESRGAN", "real_esrgan_x2.xml") have you already tested, including quantized versions, dynamic shapes, including inference on NPU?
Have you already tested dynamic video resolutions (e.g. TV broadcasting often changes resolutions between the movie and advertisement)?
Are these models already contained in/supported by the `model_downloads` scripts?

PR #1002, opened 2026-07-23
Add backend registry to gvagenai element + add openai-http backend

mholowni:
### Description

New backend for gvagenai added for using openai compatible server - /chat/completions endpoint.
Refactor of gvagenai element.

N/A

### How Has This Been Tested?

Http backend tested using OVMS with Qwen3-VL-8B-Instruct-int4-ov model.
Old backend tested using local docker environment using Phi-4-multimodal model. 

### Checklist:


yunowo:
The element property list is now messed up. Let's imagine if gvagenai adds 10 more backends (litert-lm, pytorch, onnxruntime, etc) in the future, we will have 10 incompatible property sets in one element, making users confused.
I think metapublish has a better design for "backends": gvametapublishbase base element for the common code, seperate file/mqtt/kafka for different backends, and only compiles if dependencies are available. Finally a gvametapublish bin element, for choosing backends easily.
If we have no plan for adding many new backends in the future, the easiest way would be make this openai-http feature as a new independent element. Check out https://github.com/Armaggheddon/GstVlmVision

src/monolithic/gst/elements/gvagenai/gstgvagenai.cpp
```diff
@@ -492,44 +594,52 @@ static GstFlowReturn gst_gvagenai_transform_ip(GstBaseTransform *base, GstBuffer
...
+
+            genai::GenAIResult result;
             try {
-                context->inference_tensor_vector(gvagenai->prompt_string, as_video, (float)effective_fps);
+                result = backend->submit(std::move(req)).get();
             } catch (const std::exception &e) {
-                GST_ELEMENT_ERROR(gvagenai, STREAM, FAILED, ("Failed to inference tensor vector"),
+                GST_ELEMENT_ERROR(gvagenai, STREAM, FAILED, ("Failed to run backend inference"),
                                   ("Error: %s", e.what()));
                 return GST_FLOW_ERROR;
```
msmiatac: `return GST_FLOW_ERROR;` kills the pipeline. Thus, any temporary network problem or timeout will result in the termination of the entire analytic process. The design document says we should use 'GST_ELEMENT_WARNING' and just forward the buffer without a caption.

yunowo (changes requested):
See comments above

src/monolithic/gst/elements/gvagenai/gstgvagenai.cpp
mholowni: changed in https://github.com/open-edge-platform/dlstreamer/pull/1002/commits/3eeb5be1baf5bbb665e4beb8bf6ce61beeb5b2a9

msmiatac (dismissed):

msmiatac (dismissed):

tbujewsk (changes requested):
Update docs/user-guide/elements/gvagenai.md

yunowo:
Overall, this is not good design and not scalable. ov/curl are dynamically linked to gvagenai at compile time so they are all loaded to memory when startup, but only one backend will be used. This issue will become more serious when more backends are added.

mholowni:
> Update docs/user-guide/elements/gvagenai.md

Updated

yunowo (dismissed):

dmichalo (dismissed):

msmiatac (approved):

tbujewsk (approved):

PR #1014, opened 2026-07-27
Fixed issue with resources management.

jmotow:
Fixed issue with resources management for realsense camera.


tbujewsk (dismissed):

mholowni (dismissed):

dmichalo (dismissed):

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
```diff
@@ -200,12 +200,6 @@ static void gst_real_sense_init(GstRealSense *src) {
     }
 
     gst_base_src_set_format(GST_BASE_SRC(src), GST_FORMAT_TIME);
-
-    if (detectRealSenseDevices(detectedDevices)) {
```
OskarFiedot, line 204: detectedDevices var is now unused. I think it would be good to delete it

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
```diff
@@ -264,6 +264,11 @@ static gboolean gst_real_sense_set_camera(GstRealSense *src, const gchar *camera
     }
 
     try {
+        if (src->rsPipeline) {
+            src->rsPipeline->stop();
+            src->rsPipeline.reset();
+        }
+
         src->rsPipeline = std::make_unique<rs2::pipeline>();
         if (src->rsPipeline == nullptr) {
```
OskarFiedot: Minor: `std::make_unique` never returns `nullptr` — it throws `std::bad_alloc` on failure, so this branch can't actually be reached. The surrounding try/catch already covers allocation errors, so we can safely drop this null check.

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
```diff
@@ -489,6 +494,16 @@ static GstFlowReturn gst_real_sense_create(GstBaseSrc *basesrc, guint64 offset,
 static gboolean gst_real_sense_start(GstBaseSrc *basesrc) {
     auto *src = GST_REAL_SENSE(basesrc);
 
+    if (!src->uri || !*src->uri) {
+        GST_ERROR("gst_real_sense_start: Camera device is not configured\n");
+        return FALSE;
+    }
+
+    if (!gst_real_sense_set_camera(src, src->uri, NULL)) {
```
OskarFiedot, line 534: Just want to confirm the intended behavior here: `uri` is passed into `set_camera`, but inside that function it's only null-checked and never used to select a specific device (there's no `conf.enable_device(serial)`), so the pipeline always opens the default RealSense camera. Is per-device selection by serial still something we want to support, or is single-camera the expected use case for now? If it's the former, we should wire `uri` into the config; if the latter, maybe worth a short comment so it's not mistaken for a bug later.

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
jmotow: ok

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
jmotow: Good catch - my old habit....

src/monolithic/gst/elements/gvarealsense/gvarealsense.cpp
jmotow: Right, intention was to use only uri, but you are right it was confusing. Since you raised that support for serial has been added. Thanks for pointing!

OskarFiedot (dismissed):

dmichalo (dismissed):

dmichalo (approved):

OskarFiedot (approved):

PR #1024, opened 2026-08-04
Add RF-DETR , RF-DETR-SEG converter implementation

tbujewsk:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/rfdetr.cpp
```diff
@@ -0,0 +1,176 @@
...
+            parseOutputBlobs(
+                reinterpret_cast<const float *>(logits_blob->GetData()) + logits_unbatched_size * batch_number,
+                logits_blob->GetDims(),
+                reinterpret_cast<const float *>(boxes_blob->GetData()) + boxes_unbatched_size * batch_number,
+                boxes_blob->GetDims(), objects);
+        }
+
+        return storeObjects(objects_table);
+    } catch (const std::exception &e) {
+        throw std::runtime_error("Failed to do RF-DETR post-processing.");
```
msmiatac: Use std::throw_with_nested here for parity with rtdetr.cpp (loses root cause otherwise).

This replaces the in-flight exception with a generic message and discards e, so the actual cause thrown by parseOutputBlobs (e.g. "RF-DETR logits and boxes have different query counts.", zero classes, nullptr blob, wrong precision) never reaches the caller/logs. rtdetr.cpp uses std::throw_with_nested in this exact spot, which keeps the high-level context and preserves the original exception for rethrow_if_nested unwrapping. As written, e is also bound but unused.

The nested chain is actively consumed. Utils::createNestedErrorMsg() (utils.cpp:33) recursively unwraps via std::rethrow_if_nested to build a full multi-line message.

src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/rfdetr.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+    const auto &model_input_image_info = getModelInputImageInfo();
+    const float input_width = static_cast<float>(model_input_image_info.width);
+    const float input_height = static_cast<float>(model_input_image_info.height);
+
+    for (size_t i = 0; i < num_queries; ++i) {
+        const float *logits = logits_data + i * logits_classes;
+        const float *box = boxes_data + i * boxes_object_size;
+
+        auto probs = softmax(logits, logits_classes);
```
msmiatac: **RF-DETR uses sigmoid, not softmax - and has no background class. The current post-processing applies the wrong activation and skips a valid class.**

The exported IR documents its own post-processing recipe in rt_info (written by export_openvino.py in milestone-hafnia/trainer-object-detection). The relevant entries baked into the .xml are:

- postprocessing/labels/type = "class_logits" — the labels/logits output is raw logits.
- postprocessing/labels/activation = "sigmoid" — confidences are obtained with sigmoid, per class.
- postprocessing/dets/box_format = "cxcywh_normalized" — boxes are cxcywh in [0,1].

Against that, parseOutputBlobs has two problems:

**Activation is wrong**. The code calls softmax(logits, logits_classes), but the model expects a per-class sigmoid. Softmax normalizes across classes and couples them, so the value you compare to confidence_threshold isn't the model's calibrated confidence — this shifts the operating point and can suppress/drop valid detections. Replace the softmax with an element-wise sigmoid on each logit.

**There is no background class to skip**. The code assumes index 0 is background (best_class = 1; for (c = 2; ...)), but the recipe describes plain per-class logits with no no-object slot. RF-DETR's reference post-processing scores every class (sigmoid(logits), then label = index % num_classes) and does not skip index 0. As written you drop class 0 entirely (and may misalign labels). Start the scan at 0 (best_class = 0; for (c = 1; ...)) and change the valid_classes < 2 guard to < 1.

Sketch:
```
// RF-DETR: per-class sigmoid, no background class (model rt_info: labels.activation = "sigmoid")
size_t best_class = 0;
float best_score = sigmoid(logits[0]);
for (size_t c = 1; c < valid_classes; ++c) {
    const float score = sigmoid(logits[c]);
    if (score > best_score) { best_score = score; best_class = c; }
}
```

msmiatac (changes requested):

msmiatac (approved):

jmotow (approved):

src/monolithic/gst/inference_elements/common/post_processor/converters/to_roi/rfdetr.cpp
tbujewsk: Fixed

PR #1031, opened 2026-08-06
Initial coexistence benchmark for DL Streamer and DeepStream LPR.

jmotow:
Subject:
Initial coexistence benchmark for DL Streamer and DeepStream LPR

Body:
Add the first version of a coexistence benchmark that evaluates concurrent LPR workload capacity across Intel and NVIDIA paths.

Includes:

Unified benchmark entrypoint with platform auto-detection (Intel GPU/NPU/CPU, NVIDIA GPU)
Docker-based execution flow for both inference stacks
Automatic first-run model/bootstrap setup for DL Streamer and DeepStream TAO assets
Stream-scaling benchmark loop with FPS-threshold stopping logic
Fixed-stream execution mode, including parallel dual-platform runs
Runtime diagnostics for startup failures, OOM detection, and preroll stalls
Per-platform FPS collection and summarized benchmark results
Updated documentation and usage examples for the new benchmark workflow


tbujewsk (changes requested):
This is script is too long, too complicated
1. Remove - Fixed mode at all and leave only the Benchmark mode (per design)
2. Move as much as possible methods into utils.sh e.g. like live FPS monitor

tbujewsk (dismissed):
Implement discussed changes.

tbujewsk (approved):

msmiatac (approved):

PR #1093, opened 2026-08-28
Fix PaddleOCR padding and YOLO YAML parsing

msmiatac:
### Description

This PR fixes two bugs related to model configuration reading.

1. For PaddleOCR v5 the padding value was not set explicitly, which resulted in the use of the default 0 value during padding performed by the preprocessor. Because the preprocessor applies padding **before** the normalization, the padding values were transformed to -1, although the model expects zero-padded input.
2. For YOLO models the metadata.yaml reader contained a bug in the loop for reading lists. As a result the line after a list was silently ignored.

### DIFF
--- a/src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
+++ b/src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
@@ -108,7 +108,7 @@ bool yaml2Json(const std::string yaml_file, nlohmann::json &yaml_json) {
 
                             size_t item_first = line.find_first_not_of(" \t");
                             if (item_first == std::string::npos || line[item_first] != '-') {
-                                yaml_stream.seekg(pos); // ✅ Rewind - put the line back
+                                yaml_stream.seekg(pos); // Rewind - put the line back
                                 break;
                             }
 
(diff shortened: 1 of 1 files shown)


tbujewsk (dismissed):

src/monolithic/inference_backend/image_inference/openvino/model_api_converters.cpp
```diff
@@ -100,10 +100,17 @@ bool yaml2Json(const std::string yaml_file, nlohmann::json &yaml_json) {
...
+                        while (yaml_stream.good()) {
+                            // Peek at next line without consuming it
+                            std::streampos pos = yaml_stream.tellg(); // Save position
+                            if (!std::getline(yaml_stream, line))
+                                break;
+
                             size_t item_first = line.find_first_not_of(" \t");
-                            if (item_first == std::string::npos || line[item_first] != '-')
+                            if (item_first == std::string::npos || line[item_first] != '-') {
+                                yaml_stream.seekg(pos); // ✅ Rewind - put the line back
```
dmichalo: Do we want to keep this command with emoji?

dmichalo (approved):

tbujewsk (approved):

PR #1110, opened 2026-09-07
Ubuntu 26 support

mholowni:
### Description

Add new dockerfile based on ubuntu26.
Add building of u26 packages

### Checklist:


.github/workflows/dls-pr-workflow.yaml
nszczygl9, line 404: I would suggest extending this step to build dls from sources on u26

